Issue #6, opened 2020-09-19
hierarchical kg vainishes page


cbizon:
http://robokop.renci.org:7080/question/jgJoDEYW

maximusunc:
Due to new message formats, not all nodes were getting a "level" property and that was crashing the page. Fixed in #8

PR #8, opened 2020-09-22
adding level to all kg nodes for hierarchy graph view

maximusunc:
#6 fix bug where some nodes don't get assigned a "level" and the page crashes when trying to activate hierarchical view.

### DIFF
--- a/src/stores/useMessageStore.js
+++ b/src/stores/useMessageStore.js
@@ -113,189 +113,191 @@ export default function useMessageStore() {
     return message.current.knowledge_graph.nodes[idToIndMaps.kgNodeMap.get(nodeId)];
   }
 
+  function getQgNodeInd(qgId) {
+    return idToIndMaps.qgNodeMap.get(qgId);
+  }
+
   function annotatedPrunedKnowledgeGraph(pruneNum) {
-    if (message.current.query_graph) {
-      // KG nodes don't always have type
-      // If they don't we need to figure out which qNodes they most like correspond to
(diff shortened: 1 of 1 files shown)


patrickkwang:
It looks like there was already a "level", and you're changing how it's determined. Can you explain more about what this changes does and why?

maximusunc:
To use the hierarchical view, each node in the graph needs to have a "level" property. In old Robokop message format, it looks like all nodes either had a type array or a label with no type, so as far as I can tell we were always going into the if block at the bottom of this annotatedPrunedKnowledgeGraph function. That message format is no longer the case, and we need to handle level differently. So I moved the level property logic up to where we're already mapping over all kg nodes and results and I'm finding the query graph index that each kg node corresponds to.

This could potentially prove unintuitive if one kg node corresponds with multiple qg nodes in different results, but the kg node will at least have one "level" and the viewable graph might just look weird.

patrickkwang (commented):
Miscellaneous comment. Can we flip around the if statement at line 117 so that we abort if the query graph is empty, and unindent everything by two spaces?

More general miscellaneous comment. This seems like a lot of work. Maybe it just looks like a lot and can be made more readable, or maybe it's over-complicated just to enable smart pruning. Either way this method may be a place to spend some more time.

maximusunc:
Just opened an issue for improving that function, because I agree that it can probably be better.

patrickkwang (approved):

alongreyber (approved):

PR #11, opened 2020-09-23
Added "npm run hot-reload" for optional hot reloading support

alongreyber:
Added support for hot reloading. This allows changes to React components without a full re-render or page refresh. This is useful, especially for UI development because it decreases cycle time for feedback on code changes.

The hot reload support is completely optional. `npm run start` works the same way as it did before the changes. All added dependencies are only loaded if hot reloading is enabled, meaning that there is no overhead added to either the development or production output. Hot reloading is controlled via an environment variable.


server.js
```diff
@@ -9,7 +9,21 @@ const app = express();
 app.use(bodyParser.urlencoded({ extended: true }));
 app.use(bodyParser.json({ limit: '4000mb' }));
 
-app.use(express.static('pack'));
+if(process.env.HOT_RELOAD) {
```
maximusunc, line 12: the "if" should have a space after it.

server.js
```diff
@@ -9,7 +9,21 @@ const app = express();
...
+  console.log("Hot reloading");
+  const webpack = require('webpack');
+  const webpackConfig = require('./webpack.refresh.js');
+  const compiler = webpack(webpackConfig);
+
+  app.use(require("webpack-dev-middleware")(compiler, {
+      noInfo: true, publicPath: webpackConfig.output.publicPath
```
maximusunc, line 19: There's an extra indentation here.

server.js
patrickkwang: If yall want to get crazy, you could try adding some automated linting on GitHub.

maximusunc (approved):
I'm a bit wary of the webpack plugin and think it might start doing some weird things down the line (just something we should keep an eye on), but we can certainly use this functionality in the mean time!

PR #14, opened 2020-09-25
Robokache integration API refactoring

alongreyber:
Refactored the way that the Robokop UI components interact with Robokache. Previously, this was done through the express server. Now the interaction is done directly to reduce boilerplate and increase performance.

The new API automatically parses JSON responses and always writes JSON requests. This was updated to match version 4.0 of the Robokache API (not yet merged, but part of [this PR](https://github.com/NCATS-Gamma/robokache/pull/39). All error responses include `status` and `message` keys. It is easy to check if a response failed by checking `response.status == "error"`. The message key is set when there is an error and includes a human readable error message that could be displayed on the page.

Robokache treats both questions and answers identically. The API exposed to the UI in this PR does not do this. Methods are exposed for both getQuestion and getAnswer. These methods behave identically so they point to the same function to reduce duplication. The idea behind this design choice is to make UI code more readable by always including whether it is a question or answer. Not all routes are the same between questions and answers, for example the `getAnswerByQ` route does not have an equivalent for questions. In addition, there is no `setAnswerData` method because this is likely functionality that will live in the express server.


src/API/index.js
```diff
@@ -1,87 +1,91 @@
...
+    method: method,
+    body: body && JSON.stringify(body),
+    credentials: "include",
+    headers: {
+      'Accept': 'application/json',
+      'Content-Type': 'application/json',
+    },
+  };
+  if(auth)
+    config.headers.Authorization = `Bearer ${auth}`;
```
maximusunc, line 23: I think there should be two ways to write an IF statement, a one-liner or my preferred option, with {}. I think this is just hard to read.
Also, please put a space after the if. I'm not a fan of auto-linting, but we might could look into some more specific linting rules.

src/API/index.js
```diff
@@ -1,87 +1,91 @@
...
-        reject(errorText);
-      });
-  }),
-};
+// Some of the API routes have the same method signatures for questions and answers.
+// 
+// It makes sense to expose these methods separately 
+// so when they are called in UI code it is clear
+// whether the result will be an answer or question
+routes = {...routes,
```
maximusunc, line 82: I would put this spread operator on its own line.

src/API/index.js
```diff
@@ -1,87 +1,91 @@
...
+
+let routes = {
+  async getQuestions(token) {
+    return baseRequest('document?has_parent=false', 'GET', null, token);
+  },
+
+  async getQuestion(question_id, token) {
+    return baseRequest(`document/${question_id}`, 'GET', null, token);
+  },
+  async getAnswersByQ(question_id, token) {
```
maximusunc, line 37: I think this function name can be the whole 'getAnswersByQuestion'.

src/API/index.js
```diff
@@ -1,87 +1,91 @@
...
+    return baseRequest(`document/${question_id}/children`, 'GET', null, token);
+  },
+
+  async getQuestionData(question_id, token) {
+	  let config = {
```
maximusunc, line 42: Please fix these indentations

src/API/index.js
```diff
@@ -1,87 +1,91 @@
...
-      .then((res) => {
-        resolve(res);
-      })
-      .catch((err) => {
-        const errorText = errorHandling(err);
-        reject(errorText);
-      });
-  }),
-};
+// Some of the API routes have the same method signatures for questions and answers.
```
maximusunc, line 77: If we're going to reuse these methods for both answers and questions, I would rather give the methods generic names like getDocument, createDocument, because we'll be exposing better method names down below. Then we won't have the getAnswer: routes.getQuestion, which is weird at a glance. You would need to update the exposed UI methods that are currently in the spread operator.

maximusunc (commented):
I'm still not convinced that we want to move all of these calls to the frontend UI code. It might make things slightly faster to not pass things through the express server, but I think those gains are negligible. I think it makes more sense that the UI only calls one thing, the express server, who then routes things either to Robokache or out to ARAs and other services. That seems like a better separation of concerns. I think this would also provide us with more utility down the line, like for example, if the user asks a question and wants to store it in the DB, the express server could receive that request and then go to an external service to "augment" or fill in the blank before actually storing the question. We wouldn't be able to do this if the UI were trying to directly talk with Robokache. I like this cleaner code, and we could definitely move this logic to the express server.

alongreyber:
I think it makes more sense for all of the business logic to be in the front-end, and have the back-end expose multiple services with primitive operations to build those on. For example, an operation to create a question and ask an ARA to answer that question could look like this in pseudocode:

```
let question_id = await API.cache.createQuestion(content);
API.express.askARA(question_id); // Express server contacts robokache and uses question stored there for ARA query
```

I can understand the advantages of having the logic in the express server. It makes the front-end API calls simpler. However, the simpler method signatures hide side effects that aren't immediately clear. For example, in a call to a route like `/api/createQuestion`, the express server would presumably also contact an ARA for an answer, which is not immediately clear from the method signature.


Just wanted to share my reasoning for the API structuring like this in the PR. Let me know which direction you would like me to move forward. I'd be happy to re-introduce the routes into the express server if you think that's the best way to go.

maximusunc:
Ok, let's keep all routing logic in the UI code. The express server will be "dumb". The Robokop UI will only be in charge of question upload and answer getting, and everything outside of that will be the responsibility of the ARAs.

Please convert your fetch calls to axios.

maximusunc (changes requested):
Please convert your fetch calls to axios.
Please pull main branch.
Please update the existing api calls in the ui code to use this API.

src/API/index.js
```diff
@@ -1,87 +1,99 @@
...
-  getQuestions: (token) => new Promise((resolve, reject) => {
-    const config = {
-      method: 'get',
-      url: '/api/questions',
-      headers: {},
-    };
-    if (token) {
-      config.headers.authorization = `Bearer ${token}`;
+    if(token)
+      config.headers.Authorization = `Bearer ${token}`;
```
maximusunc, line 49: Please fix this syntax. Also make this variable and line 23 consistent. 'Token' makes more sense to me.

src/API/index.js
```diff
@@ -1,87 +1,121 @@
-import axios from 'axios';
+const axios = require('axios');
```
maximusunc, line 1: Can you make this an import instead of a require?

src/API/index.js
```diff
@@ -1,87 +1,121 @@
...
+/**
+ * URL Maker
+ * @param {string} ext extension to append to url
+ */
+const base_url = `${protocol}://${host}:${port}/api/`;
+
+function handleAxiosError(error) {
+	let errorResponse;
+	if (error.response) {
+		errorResponse = error.response.data;
```
maximusunc, line 16: Please fix all indentations.

src/API/index.js
```diff
@@ -1,87 +1,121 @@
...
+		errorResponse.status = "error";
```
maximusunc, line 17: You're setting errorResponse to a string in the line right above this.

src/API/index.js
```diff
@@ -1,87 +1,121 @@
...
+		headers: {
+			'Accept': 'application/json',
+			'Content-Type': 'application/json',
+		},
+	};
+	if (token) {
+		config.headers.Authorization = `Bearer ${token}`;
+	}
+
+	let response = await axios(config).catch(handleAxiosError);
```
maximusunc, line 45: I'll admit, I'm used to writing promises with then/catch. But we can switch over to async/await. But I would suggest we stick with one idiom and not mix. This should probably be in a try/catch block?

maximusunc (changes requested):
There are also all the axios calls in Answers.jsx which need to be converted to use this API.

src/API/index.js
alongreyber: `error.response.data` is a JSON object. I'm setting the status property of that object so that it's easy to manage error handling in the UI with `response.status == 'error'`.

src/pages/Answers.jsx
```diff
@@ -24,80 +25,52 @@ export default function Answer({ user }) {
...
-              setErrorMessage(`Unable to retrieve answers for question: ${question_id}`);
-              toggleLoading(false);
-            });
-        })
-        .catch((err) => {
-          console.log('err', err);
-          setErrorMessage('Unable to load question.');
-          toggleLoading(false);
-        });
+  let token;
```
maximusunc, line 28: Let's move this into the fetchQuestionAnswerData function, because this will otherwise be run on every rerender.

src/pages/Answers.jsx
```diff
@@ -24,80 +25,52 @@ export default function Answer({ user }) {
...
+
+  async function fetchQuestionAnswerData() {
+    let response;
+    response = await API.getQuestionData(question_id, token);
+    if (response.status == 'error') {
+      setErrorMessage('Unable to load question.');
+      toggleLoading(false);
+      return;
+    }
+    let query_graph = response;
```
maximusunc, line 41: Use 'const' here and on line 49. We seem to have lost our eslint file when we moved the ui folder up (I'll make a separate PR to add that back in). 'let' should only be used if you are planning on overwriting the variable's value.

src/pages/Answers.jsx
```diff
@@ -24,80 +25,52 @@ export default function Answer({ user }) {
...
+    }
+
+    const { knowledge_graph, results } = response;
+    const message = { query_graph, knowledge_graph, results };
+    messageStore.initializeMessage(message);
+    setMessageSaved(true);
+    toggleLoading(false);
+  }
+
+  useEffect(() => { fetchQuestionAnswerData() },
```
maximusunc, line 72: Please change this format to the same as what's commented out right below.

src/pages/SimpleViewer.jsx
```diff
@@ -27,64 +28,50 @@ export default function SimpleViewer(props) {
...
-          headers: {
-            'Content-Type': 'multipart/form-data',
-            Authorization: `Bearer ${user.id_token}`,
-          },
-        })
-          .then((res) => {
-            // console.log('question updated', res.data);
-            const answerData = new FormData();
-            const answerBlob = new Blob([JSON.stringify({
+  console.log("Rendering SimpleViewer");
```
maximusunc, line 31: Please remove this console.log.

src/pages/SimpleViewer.jsx
```diff
@@ -27,64 +28,50 @@ export default function SimpleViewer(props) {
...
+    // Create Answer
+    response = await API.createAnswer({parent: questionId, visibility: 1}, user.id_token);
+    if (response.status == 'error') {
+      setErrorMessage('Unable to create answer object.');
+      return;
+    }
+    const answerId = response.id;
+    // Upload answer data
+    const answerData = JSON.stringify({
               knowledge_graph: messageStore.message.knowledge_graph,
```
maximusunc, line 64: Please fix this indentation

maximusunc (changes requested):

maximusunc (approved):

PR #15, opened 2020-09-25
removing a bunch of unused code

maximusunc:


patrickkwang (dismissed):
I trust you. And we can always fish out bits of this later.

alongreyber (approved):

PR #22, opened 2020-09-29
Adding eslintrc back in and adding a readme

maximusunc:


package.json
```diff
@@ -11,29 +11,30 @@
...
     "file-loader": "^6.0.0",
     "html-webpack-plugin": "^4.3.0",
     "nodemon": "^2.0.4",
+    "react-refresh": "^0.8.3",
     "style-loader": "^1.2.1",
     "url-loader": "^4.1.0",
     "webpack": "^4.43.0",
     "webpack-cli": "^3.3.12",
-    "webpack-merge": "^5.0.9",
-    "@pmmmwh/react-refresh-webpack-plugin": "^0.4.2",
```
alongreyber, line 32: It looks like removing this breaks hot reloading.

alongreyber (changes requested):

package.json
maximusunc: hmm, I didn't remove it, it's just moved up because npm installing auto-generates this list.

alongreyber: Yeah just noticed that too

alongreyber (approved):

PR #37, opened 2020-10-05
Messenger ara

maximusunc:


apiRoutes/routes.js
```diff
@@ -0,0 +1,51 @@
+const router = require('express').Router();
+const axios = require('axios');
+
+const robokache = require('../src/API/robokache');
```
alongreyber, line 4: I don't like the idea of sharing code between the frontend and backend. My view is that they are two distinct entities and are running in separate environments. Seems like we could run into weird errors if, for example, we tried to access `window` somewhere in the robokache source file. In addition, we have to share all dependenices between the frontend and backend which bloats the runtime environment of the backend and the bundle size of the frontend. It also just feels like bad practice to so tightly couple the components. 

We can't separate these now, but maybe we should think about trying not to share code so that we have more flexibility in the future. What that would look like is keeping everything in apiRoutes self contained. Right now that would involve copying the src/API/ara.js file into apiRoutes as well as duplicating some of src/API/robokache.js in apiRoutes.

apiRoutes/routes.js
```diff
@@ -0,0 +1,51 @@
...
+const handleAxiosError = require('../src/API/utils');
+
+const messenger = 'http://robokop.renci.org:4868';
+
+router.route('/answer')
+  .post(async (req, res) => {
+    const { questionId } = req.query;
+    try {
+      // TODO: do we first check to see if the question is owned by user?
```
alongreyber, line 13: We shouldn't need to check this. Answer ownership is managed by robokache.

src/pages/SimpleViewer.jsx
```diff
@@ -28,6 +29,48 @@ export default function SimpleViewer(props) {
   const [showSnackbar, toggleSnackbar] = useState(false);
   const messageStore = useMessageStore();
 
+  async function askQuestion() {
+    const defaultQuestion = {
+      parent: '',
+      visibility: 1,
+      metadata: { name: `Simple Viewer Question - ${formatDateTimeNicely(new Date().toLocaleString())}` },
```
alongreyber, line 36: Not sure we need the date in the name. The date is in the column right next to the name in QuestionList so its just duplicate information and makes the table more messy.

src/API/server.js
```diff
@@ -0,0 +1,34 @@
+import axios from 'axios';
```
alongreyber, line 1: I think we need a better name for the express server. Right now we have a `server.js` which is not a descriptive name and could be confusing. We also have a folder called `apiRoutes` that is also confusing because there are many APIs that we are exposing/interacting with. Overall, the terms API and server are very overloaded in this codebase.

Here are a couple names I thought of for the express server that I think could work:

- TaskQueue
- AnswerQueue
- AnswerDispatcher
- QueryDispatcher
- ReasonerQueryDispatcher

I like ReasonerQueryDispatcher the best. It's long, but at least it says exactly what is happening. If you want to go with that we should rename API/server.js -> API/reasonerQueryDispatcher.js and apiRoutes/ -> reasonerQueryDispatcherServer/. API calls would look something like:

```
API.reasonerQueryDispatcher.getAnswer(questionId)
```

src/pages/SimpleViewer.jsx
```diff
@@ -28,6 +29,48 @@ export default function SimpleViewer(props) {
   const [showSnackbar, toggleSnackbar] = useState(false);
   const messageStore = useMessageStore();
 
+  async function askQuestion() {
+    const defaultQuestion = {
+      parent: '',
+      visibility: 1,
```
alongreyber, line 35: I think we should avoid using the visibility numbers. In utils/cache.js there is a mapping of numbers to strings for visibility. We should probably generate an inverse of that mapping and use that instead of the constant numbers.

src/pages/SimpleViewer.jsx
```diff
@@ -127,7 +169,10 @@ export default function SimpleViewer(props) {
                 omitHeader
               />
               {user && (
-                <button type="button" onClick={uploadMessage}>Upload</button>
+                <>
+                  <button type="button" onClick={uploadMessage}>Upload</button>
```
alongreyber, line 173: We should use the Material UI buttons.

src/pages/questionAnswerViewer/QuestionAnswerViewer.jsx
```diff
@@ -136,6 +155,13 @@ export default function QuestionAnswerViewer() {
                     )) }
                   </Select>
                 </FormControl>
+                {user && user.id_token && (
```
alongreyber, line 158: This button is currently only rendered when there are existing answers. Might be worth it to move it up so that is displayed even if there are no answers, so that a user can manually trigger an answer lookup if something went wrong when submitting the first time.

src/pages/questionAnswerViewer/QuestionAnswerViewer.jsx
```diff
@@ -80,12 +83,28 @@ export default function QuestionAnswerViewer() {
     return answers.find((a) => a.id === id);
   }
 
+  useEffect(() => {
+    fetchAnswers();
+  }, [user, question_id, answer_id]);
+
+  async function getNewAnswer() {
+    // TODO: do we only let the owner of a question ask for new answer?
```
alongreyber, line 91: We should only render the getAnswer button if `question.owned` is true.

alongreyber (changes requested):

apiRoutes/routes.js
maximusunc: I think this needs to be handled somewhere outside of Robokache. A signed in user could go to a non-owned public question and try and ask an ara. Then we have a user owning an answer to someone else's question, which I think is a bad idea.

src/pages/SimpleViewer.jsx
maximusunc: Good point.

src/API/server.js
maximusunc: I don't think we necessarily need Reasoner in the name. But I think QueryDispatcher is a very good descriptive name.

apiRoutes/routes.js
maximusunc: This ownership logic will be handled by the UI, so the server doesn't have to worry about it.

alongreyber (changes requested):

src/API/ara.js
```diff
@@ -0,0 +1,34 @@
+import axios from 'axios';
```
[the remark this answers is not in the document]
maximusunc: Right, so I put this in early. Once we get the new question page back up, a simple question will directly talk to the ara from the frontend.

src/pages/SimpleViewer.jsx
```diff
@@ -127,7 +170,10 @@ export default function SimpleViewer(props) {
                 omitHeader
               />
               {user && (
-                <button type="button" onClick={uploadMessage}>Upload</button>
+                <>
+                  <Button onClick={uploadMessage}>Upload</Button>
```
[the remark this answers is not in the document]
maximusunc: Filled is fine. I intentionally haven't put any effort in styling these because I added them for development purposes, and they will most likely move somewhere else once we have a more fleshed-out UX design.

src/pages/SimpleViewer.jsx
```diff
@@ -127,7 +153,21 @@ export default function SimpleViewer(props) {
...
+                    variant="contained"
+                    style={{ marginRight: 10 }}
+                  >
+                    Upload
+                  </Button>
+                  <Button
+                    onClick={askQuestion}
+                    variant="contained"
+                  >
+                    Ask ARA
```
alongreyber, line 168: I tried the Ask ARA functionality with the covidkop response. No results or knowledge_graph were returned and the error message was "Unable to parse message". I think we should change that error to be more descriptive when there are no answers returned. My thinking is that we parse those messages without error and then display an error message in the AnswersetView.

src/pages/questionAnswerViewer/QuestionAnswerViewer.jsx
```diff
@@ -106,7 +109,32 @@ export default function QuestionAnswerViewer() {
...
+  }
+
+  useEffect(() => {
+    fetchAnswers();
+  }, [user, question_id, answer_id]);
+
+  async function getNewAnswer() {
+    const response = await API.queryDispatcher.getAnswer(question_id, user.id_token);
+    if (response.status === 'error') {
+      pageStatus.setFailure(response.message);
```
alongreyber, line 128: This error response doesn't have a message property so nothing happens when we hit this line. It looks like there's a `detail` property with some kind of structured logging. Not sure if we have a component to display that error. If not, we could just put a static error message.

src/pages/SimpleViewer.jsx
```diff
@@ -136,8 +176,7 @@ export default function SimpleViewer(props) {
                 message="Message saved successfully!"
```
alongreyber, line 176: We should replace this snackbar with the alert context for visual consistency.

src/pages/questionAnswerViewer/QuestionAnswerViewer.jsx
```diff
@@ -106,7 +109,32 @@ export default function QuestionAnswerViewer() {
...
+      return;
+    }
+    displayAlert(
+      'success',
+      <Button onClick={() => history.replace(`/question/${question_id}/answer/${response.id}`)}>
+        A new answer is ready!
```
alongreyber, line 134: It's not fully clear that this is a button. Maybe we should have two lines, first line says "A new answer is ready!" and the second line is a button that says "Go" and takes you to the answer.

queryDispatcher/robokache.js
```diff
@@ -0,0 +1,80 @@
+const axios = require('axios');
+const handleAxiosError = require('./utils');
+
+const protocol = 'http';
+const host = process.env.ROBOKACHE_HOST || 'lvh.me';
+const port = 8080;
+
+/**
+ * URL Maker
```
alongreyber, line 9: This comment isn't accurate anymore because this isn't a function.

alongreyber (changes requested):

src/pages/SimpleViewer.jsx
maximusunc: Opened issue #40

src/pages/questionAnswerViewer/QuestionAnswerViewer.jsx
maximusunc: I put in a check for if 'message' is missing from an error response, but we should think more on this as we add new ARAs with potentially different error reporting. #41

src/pages/SimpleViewer.jsx
```diff
@@ -1,62 +1,99 @@
...
-  // const [concepts, setConcepts] = useState([]);
-  const [errorMessage, setErrorMessage] = useState('');
-  const [loadingMessage, setLoadingMessage] = useState('');
-  const [showSnackbar, toggleSnackbar] = useState(false);
   const messageStore = useMessageStore();
+  const pageStatus = usePageStatus(false);
+  const visibility = useVisibility();
+  const displayAlert = useContext(AlertContext);
+
+  async function askQuestion() {
```
alongreyber, line 26: I think that all of the error handling in askQuestion and uploadMessage should display alerts instead of setting the pageStatus to failed. The whole page is still functional, so we should just let the user know something went wrong and let them continue using the page.

alongreyber (approved):
Looks good to me! After making the one change feel free to merge.

alongreyber (approved):

PR #44, opened 2020-10-14
SimpleQuestion builder functionality reimplementation

alongreyber:
Fixed SimpleQuestion to work with building and submitting questions from scratch. There is still some functionality that needs to be re-implemented, including adding constraints to nodes and edges, but the core functionality works.


src/pages/question/panels/NodePanel.jsx
```diff
@@ -128,7 +132,7 @@ export default function NodePanel({ panelStore }) {
           />
           {!showOptions && node.curie.length > 0 && (
             <InputGroup.Addon>
-              {node.curie[0]}
+              {node.curie}
```
maximusunc, line 135: Curie can be an array.

src/pages/question/questionBuilder/useEdge.js
```diff
@@ -29,21 +23,45 @@ export default function useEdgePanels() {
...
-    setDisablePredicates(false);
   }
 
   function initialize(seed) {
-    console.log(seed);
     reset();
-    setId(seed.id || '');
+    setId(seed.id || null);
     setSourceId(seed.source_id || null);
+    if (seed.target_id !== null) {
```
maximusunc, line 32: I think line 31 is a cleaner way of doing this.

src/pages/question/questionBuilder/useNewQuestionPanel.js
```diff
@@ -32,18 +34,22 @@ export default function useNewQuestionPanel() {
    * @param {string} id unique id of node or edge. i.e. n0, n1, e0...
    */
   function openPanel(type, id) {
+    console.log(id);
```
maximusunc, line 37: please remove this.

src/pages/question/questionBuilder/useNewQuestionPanel.js
```diff
@@ -32,18 +34,22 @@ export default function useNewQuestionPanel() {
...
     setPanelType(type);
     if (type === 'node') {
-      if (id) { // load node from query graph
-        const nodeSeed = query_graph.nodes.find((n) => n.id === id);
+      // Can't do if(id) because that evaluates to false when id == 0 🤦
```
maximusunc, line 40: You can if id is a string 😛

src/pages/question/questionBuilder/useNewQuestionPanel.js
```diff
@@ -32,18 +34,22 @@ export default function useNewQuestionPanel() {
...
         node.initialize(nodeSeed);
       } else { // new node
         node.reset();
         setName(`n${query_graph.nodes.length}`);
       }
-    } else if (id) { // load edge from query graph
-      const edgeSeed = query_graph.edges.find((e) => e.id === id);
+    } else if (id !== null && id !== undefined) { // load edge from query graph
+      const edgeSeed = query_graph.edges[id];
+      console.log(edgeSeed);
```
maximusunc, line 52: please get rid of this.

src/pages/question/questionBuilder/useNewQuestionPanel.js
```diff
@@ -67,22 +73,75 @@ export default function useNewQuestionPanel() {
...
-    q_graph.nodes = q_graph.nodes.filter((n) => nodeIds.has(n.id));
+    q_graph.nodes = q_graph.nodes.filter((n, i) => nodeIds.has(i));
     return q_graph;
   }
 
   /**
    * Remove current node by active id
    */
   function removeNode() {
+    console.log(activePanelId);
```
maximusunc, line 84: no debugging console.logs in PRs.

maximusunc (changes requested):
I'm not able to create a node, then go and change it to something else. the save and undo buttons are greyed out. I haven't dug into it, but this is a needed feature. Also look into deleting nodes. Things go bad when I delete things. This might just be fixed with the node id revert, but is something you should test.

maximusunc (changes requested):

src/pages/question/questionBuilder/useNewQuestionPanel.js
```diff
@@ -26,48 +29,92 @@ export default function useNewQuestionPanel() {
...
-    q_graph.nodes = q_graph.nodes.filter((n) => nodeIds.has(n.id));
+    const floatingNodeIDs = Object.keys(q_graph.nodes).filter((id) => !(notFloatingNodeIDs.has(id)));
+
+    // Trim a node if it is floating and marked for deletion
+    const nodesToTrim = Object.entries(q_graph.nodes)
+      .filter(([id, n]) => floatingNodeIDs.includes(id) && n.deleted)
+      .map(([id, n]) => id);
+
+    q_graph.nodes = _.omitBy(q_graph.nodes, (n, id) => nodesToTrim.includes(id));
+    // q_graph.nodes = _.pick(q_graph.nodes, notFloatingNodeIDs);
```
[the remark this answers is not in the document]
maximusunc: Bump.

src/pages/question/questionBuilder/useNewQuestionPanel.js
```diff
@@ -76,24 +123,86 @@ export default function useNewQuestionPanel() {
    */
   function removeNode() {
     const q_graph = _.cloneDeep(query_graph);
-    const activeInd = q_graph.nodes.findIndex((n) => n.id === activePanelId);
-    q_graph.nodes[activeInd].deleted = true;
+    const nodeToDelete = q_graph.nodes[panelInfo.id];
+    nodeToDelete.deleted = true;
```
maximusunc, line 127: I think this can be a one-liner. No need to set a variable here.

src/pages/question/questionBuilder/useNode.js
```diff
@@ -53,21 +54,61 @@ export default function useNodePanels() {
...
+    setName(entityNameDisplay(entry.name));
+    if (entry.curie) {
+      setCurie(entry.curie);
+    }
+    setType(conceptListToString(entry.type));
   }
 
+  async function fetchCuries(newSearchTerm) {
+    // Get and update list of curies
+    const response = await API.ranker.entityLookup(newSearchTerm);
```
maximusunc, line 78: We should handle any errors here.

src/pages/question/subComponents/ButtonGroupPanel.jsx
```diff
@@ -11,10 +11,11 @@ import { FaPlusSquare, FaPlus } from 'react-icons/fa';
  * @param {function} toggleJsonEditor function to close the advanced json editor
  */
 export default function ButtonGroupPanel({ panelStore, toggleJsonEditor }) {
-  const atleastTwoNodes = panelStore.query_graph.nodes.filter((node) => !node.deleted).length >= 2;
+  const numNodes = Object.entries(panelStore.query_graph.nodes)
```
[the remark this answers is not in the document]
maximusunc: Bump.

src/utils/queryGraph.js
```diff
@@ -0,0 +1,40 @@
+/*
+   * Load the internal query_graph from a list representation
+   * where nodes and edges are lists with an internal "id" property
+   */
+function fromListRepresentation(listRepresentation) {
+  const newQueryGraph = { nodes: {}, edges: {} };
+
+  listRepresentation.nodes.forEach((node) => {
+    newQueryGraph.nodes[node.id] = { ...node, id: undefined };
```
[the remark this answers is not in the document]
maximusunc: Bump.

maximusunc (changes requested):
Here are some of the missing functionality/issues that I'm seeing:
- Edge panel header should be colored according to the source and target node types
- switching target and source nodes in editing edge panel, Save button is greyed out
- edge panel, source and target nodes should have their ids in the displayed name. Helps when you have multiple nodes of the same name/type
- edge panel, target node dropdown can have total number of predicates available.
- graph display, each node should have its id in its display name.
- question store: getListRepresentation should nest the nodes and edges in a query_graph
- graph display, after adding a second node, the graph doesn't update correctly. i.e. I can't see the second node until I zoom out or otherwise manually cause a page rerender.
- graph display, edges with no predicates shouldn't be directed. i.e. they shouldn't display an arrow going to the target node. This is happening in QuestionGraphView because the getDisplayOptions is getting an old version of the question prop not from the getDisplayGraph function.
- edge panel, source and target should be auto-filled
- edge panel, predicates look like they might not be getting cleared out when panel is closed.
- graph display, something is weird with height, if you add three floating nodes, they start going outside of the display area, but this doesn't happen in the deployed version. might be another getDisplayOptions bug.

src/pages/question/questionBuilder/QuestionBuilder.jsx
```diff
@@ -71,30 +76,48 @@ export default function QuestionBuilder(props) {
...
-        }
-      };
+      fr.onload = (e) => validateAndParseFile(e.target.result);
       fr.onerror = () => {
-        window.alert('Sorry but there was a problem uploading the file. The file may be invalid JSON.');
-        questionStore.resetQuestion();
-        setLoading(false);
-        // questionStore.setGraphState(graphStates.error);
+        displayAlert('error', 'Sorry but there was a problem uploading the file. Please try again.');
+        setStep('options');
```
maximusunc, line 120: From a code perspective, I would think we want a setLoading(false); here, but it doesn't seem to matter. So either the fr.onloadstart function is not run when I think it is, or the alert is rerendering the page. Don't know if we need to do anything, but this seems weird.

src/pages/question/questionBuilder/useNode.js
```diff
@@ -68,46 +67,36 @@ export default function useNodePanels() {
     setSearchTerm(entityNameDisplay(entry.name));
     setName(entityNameDisplay(entry.name));
     if (entry.curie) {
-      setCurie(entry.curie);
+      setCurie([entry.curie]);
```
maximusunc, line 70: Can we be certain that entry.curie will always be a string and not an array here? I think this is fine for now, but something to keep in mind.

src/pages/question/subComponents/ButtonGroupPanel.jsx
```diff
@@ -11,8 +11,8 @@ import { FaPlusSquare, FaPlus } from 'react-icons/fa';
  * @param {function} toggleJsonEditor function to close the advanced json editor
  */
 export default function ButtonGroupPanel({ panelStore, toggleJsonEditor }) {
-  const numNodes = Object.entries(panelStore.query_graph.nodes)
-    .filter(([id, node]) => !node.deleted).length;
+  const numNodes = Object.keys(panelStore.query_graph.nodes)
+    .filter((id) => !panelStore.query_graph.nodes[id].deleted).length;
```
maximusunc, line 15: Instead of keys, this can be Object.values(), which will give you each node object. The way you've done it works, but isn't the most optimal.

src/pages/question/useQuestionStore.js
```diff
@@ -25,7 +25,7 @@ export default function useQuestionStore() {
 
   function resetQuestion() {
     updateQuestionName('');
-    updateQueryGraph(null);
+    updateQueryGraph(queryGraphUtils.getEmptyQueryGraph());
```
maximusunc, line 28: This should be getEmptyGraph(). Right now this throws an error.

src/utils/queryGraph.js
```diff
@@ -1,9 +1,16 @@
...
 /*
    * Load the internal query_graph from a list representation
    * where nodes and edges are lists with an internal "id" property
    */
 function fromListRepresentation(listRepresentation) {
-  const newQueryGraph = { nodes: {}, edges: {} };
+  const newQueryGraph = getEmptyGraph();
 
   listRepresentation.nodes.forEach((node) => {
     newQueryGraph.nodes[node.id] = { ...node, id: undefined };
```
maximusunc, line 16: This is still here. We should remove id from inside the node object.

src/pages/question/panels/EdgePanel.jsx
```diff
@@ -30,16 +30,16 @@ export default function EdgePanel(props) {
...
+    const targetNode =
+      { ...panelStore.query_graph.nodes[edge.targetId] };
+      // Including name or empty curie list in the node breaks the API call
       // So we make a copy and remove it
-    sourceNode = { ...sourceNode };
-    targetNode = { ...targetNode };
     delete sourceNode.name;
     delete targetNode.name;
+    if (sourceNode.curie.length === 0) delete sourceNode.curie;
+    if (targetNode.curie.length === 0) delete targetNode.curie;
```
maximusunc, line 42: These are throwing errors when an uploaded question has curie as a string.

src/pages/question/questionBuilder/useNode.js
alongreyber: Entries come from the ranker API which says that the resulting curie will be a string. I think it makes sense to "trust" whatever API we are calling instead of handling the possibility that it is not behaving as documented.

maximusunc (changes requested):

src/pages/question/questionBuilder/useNode.js
maximusunc: 👍

src/pages/question/panels/EdgePanel.jsx
```diff
@@ -15,31 +22,76 @@ const listItem = ({ item }) => (
...
+    const sourceNode =
+      { ...panelStore.query_graph.nodes[edge.sourceId] };
+    const targetNode =
+      { ...panelStore.query_graph.nodes[edge.targetId] };
+      // Including name or empty curie list in the node breaks the API call
+      // So we make a copy and remove it
+    delete sourceNode.name;
+    delete targetNode.name;
+    if (sourceNode.curie.length === 0) delete sourceNode.curie;
+    if (targetNode.curie.length === 0) delete targetNode.curie;
```
maximusunc, line 42: Sorry. We also need to handle when a node doesn't have a curie. A node could be as simple as { type: 'gene' }.

maximusunc (changes requested):

maximusunc (changes requested):
I'm also getting "CORS" issues again when trying to ask a question. Meaning we're sending some stuff that Messenger doesn't know what to do with.

maximusunc (approved):

Issue #18, opened 2020-09-28
Find token expiration solution


maximusunc:
If a user sits on the page long enough, their auth token will expire, and then the server will respond with a 500 error. We could go and try and refresh the user's token if a 'token is expired' message comes back, or just get rid of any token expiration.

maximusunc:
#52

PR #52, opened 2020-10-16
Refresh access token

maximusunc:
I was able to leave the tab open over night and then came back in the morning and was able to continue using the app without getting a 'token expired' error.

Takes care of #18


src/components/header/Header.jsx
```diff
@@ -20,10 +20,32 @@ export default function Header({ setUser }) {
...
+  function refreshToken(timeToRefresh, googleUser) {
+    setTimeout(async () => {
+      try {
+        const { id_token, expires_in } = await googleUser.reloadAuthResponse();
+        const username = googleUser.getBasicProfile().Ad;
+        setUser({ username, id_token });
+        const newRefreshTime = (expires_in - 60) * 1000;
+        refreshToken(newRefreshTime, googleUser);
+      } catch (err) {
+        console.log('Error refreshing user:', err.type);
```
alongreyber, line 38: The header has access to the Alert context provider so let's use that here.

alongreyber (changes requested):

alongreyber:
> I was able to leave the tab open over night and then came back in the morning and was able to continue using the app without getting a 'token expired' error.
> 
> Takes care of #18

You can also test it like this so you don't have to wait all night:

Screen Shot 2020-10-16 at 12 34 20 PM

maximusunc:
Yes, while developing, I shortened the expiry time to be very short. Once I was done making code changes, I also let it sit open over night and was able to come back and everything worked without having to refresh.

alongreyber (changes requested):
There are still two console.log statements for errors. Github won't let me annotate them but I think we should switch them to alerts since we're doing that for the rest of them in this file.

alongreyber (approved):

PR #59, opened 2020-10-23
Biolink integration

alongreyber:


src/pages/question/panels/NodePanel.jsx
```diff
@@ -121,6 +123,44 @@ export default function NodePanel({ panelStore }) {
...
+
+  function updateConceptList() {
+    const spacesToSnakeCase = (str) => str.replaceAll(' ', '_').toLowerCase();
+    const setify = (name) => `Set of ${name}s`;
+
+    if (!biolink) {
+      return;
+    }
+    const validConcepts = biolinkUtils.getValidConcepts(biolink);
+    const conceptsFormatted = Object.entries(validConcepts).map(
```
maximusunc, line 145: It looks like this can be Object.keys()

src/pages/question/questionBuilder/useNode.js
```diff
@@ -82,9 +75,12 @@ export default function useNodePanels() {
...
 
   function updateFilteredConcepts(value) {
-    const newFilteredConcepts = conceptsWithSets.filter(
-      (concept) => concept.name.includes(value.toLowerCase()),
-    );
+    // Convert name to lowercase before searching
+    const newFilteredConcepts = concepts
+      .map((c) => ({ ...c, name: c.name.toLowerCase() }))
+      .filter(
+        (concept) => concept.name.includes(value.toLowerCase()),
```
maximusunc, line 82: I would change this line to (concept) => concept.name.toLowerCase().includes(value.toLowerCase()), because right now you're looping over the list twice but you can do it just once and get rid of the map on line 80.

src/utils/biolink.js
```diff
@@ -0,0 +1,36 @@
+import _ from 'lodash';
+
+const standardizeCase = (str) => str && str.replaceAll(' ', '_').toLowerCase();
```
maximusunc, line 3: This function name could be better. What is standard?

src/pages/question/panels/NodePanel.jsx
```diff
@@ -121,6 +123,44 @@ export default function NodePanel({ panelStore }) {
...
+      displayAlert('error',
+        'Failed to contact server to download biolink model. You will not be able to select concepts. Please try again later');
+      return;
+    }
+    setBiolink(response);
+  }
+  useEffect(() => { fetchBiolink(); }, []);
+
+  function updateConceptList() {
+    const spacesToSnakeCase = (str) => str.replaceAll(' ', '_').toLowerCase();
```
maximusunc, line 138: This is a redundant function coming from biolinkUtils.standardizeCase.

src/utils/biolink.js
maximusunc: All the functions in this util file could also use jsdocs.

maximusunc (changes requested):

src/pages/question/panels/EdgePanel.jsx
```diff
@@ -26,33 +28,60 @@ export default function EdgePanel(props) {
...
-      })),
+    return Object.entries(biolink.slots).map(
+      ([identifier, predicate]) => ({
+        name: biolinkUtils.standardizeCase(identifier),
+        domain: biolinkUtils.standardizeCase(predicate.domain),
+        range: biolinkUtils.standardizeCase(predicate.range),
+      }),
     );
   }
+  const predicateList = getPredicateList() || [];
```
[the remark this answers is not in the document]
alongreyber: Is this an issue? I would rather not optimize if not necessary. Putting this in a useEffect would risk us failing to re-render if we don't have a dependency set correctly. getPredicateList doesn't require an API call and the predicate list is pretty short.

src/pages/question/panels/EdgePanel.jsx
```diff
@@ -26,33 +28,60 @@ export default function EdgePanel(props) {
...
+
+    const sourceNodeTypeHierarchy = biolinkUtils.getHierarchy(biolink, sourceNode.type);
+    const targetNodeTypeHierarchy = biolinkUtils.getHierarchy(biolink, targetNode.type);
+    return predicateList.filter(
+      (p) => sourceNodeTypeHierarchy.includes(p.domain) &&
+               targetNodeTypeHierarchy.includes(p.range),
+    );
+  }
+
+  const filteredPredicateList = getFilteredPredicateList() || [];
```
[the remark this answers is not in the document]
alongreyber: Same as above comment.

src/pages/question/panels/EdgePanel.jsx
```diff
@@ -83,15 +112,30 @@ export default function EdgePanel(props) {
...
+
+  const isValid = edge.sourceId && edge.targetId && isValidPredicate;
+
   const disabledSwitch = edge.sourceId === null || edge.targetId === null;
 
+  useEffect(() => {
+  // Update edge in panelStore
+    edge.setIsValid(isValid);
+    edge.setIsValidPredicate(isValidPredicate);
+  }, [edge.sourceId, edge.targetId, edge.predicate]);
```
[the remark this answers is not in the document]
alongreyber: I agree that this is an issue. The problem is that isValid is read directly in useNewQuestionPanel so it needs to be a property of useEdge. We can't move all of the logic for isValid into useEdge because it includes API calls, which require access to alert for error handling (unavailable in hooks). 

The [refactoring proposal](https://github.com/NCATS-Gamma/robokop-rewrite/issues/45) provides a more permanent solution to this issue by requiring a single source of state and moving panel controls within the EdgePanel.

src/pages/question/panels/EdgePanel.jsx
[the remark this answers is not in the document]
alongreyber: Also, useEffect would require another useState for the predicate list.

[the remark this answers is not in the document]
maximusunc: I could maybe see this as a case for the filteredPredicateList, but the full predicateList is only dependent on the biolink object, and really only needs to be run once right after we receive the biolink from the API call.

src/pages/question/panels/EdgePanel.jsx
```diff
@@ -26,33 +28,60 @@ export default function EdgePanel(props) {
...
-      // So we make a copy and remove it
-    delete sourceNode.name;
-    delete targetNode.name;
-    if (sourceNode.curie && sourceNode.curie.length === 0) delete sourceNode.curie;
-    if (targetNode.curie && targetNode.curie.length === 0) delete targetNode.curie;
-    const response = await API.ranker.predicateLookup(sourceNode, targetNode);
-    if (response.status === 'error') {
-      predicateStatus.setFailure('Failed to contact predicate lookup server. Please try again later');
+    setBiolink(response);
+  }
```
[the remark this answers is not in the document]
alongreyber: I moved the API call into QuestionGraphViewContainer. I think having the call on application load has a couple issues:

- The API call still takes place even if the user is not going to use the question builder
- We would either need to use props everywhere or have a context object which is not recommended for something specific to a couple of components

I think that putting the API call in QuestionGraphViewContainer is a good balance. The API call happens when the question builder is rendered and only requires two props.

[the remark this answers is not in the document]
maximusunc: Perhaps I wasn't being explicit enough. The classes in the biolink model are going to replace the concepts in the config.json file right next to App.jsx, and is being used in every place we're displaying a node. This is most likely going to be another context object.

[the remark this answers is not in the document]
alongreyber: That makes sense. I've switched it to a context object.

src/pages/question/panels/NodePanel.jsx
```diff
@@ -13,11 +13,14 @@ import shortid from 'shortid';
 import AlertContext from '@/context/alert';
 import API from '@/API';
 
-import entityNameDisplay from '../../../utils/entityNameDisplay';
-import curieUrls from '../../../utils/curieUrls';
+import entityNameDisplay from '@/utils/entityNameDisplay';
+import curieUrls from '@/utils/curieUrls';
+import biolinkUtils from '@/utils/biolink';
 import getNodeTypeColorMap from '../../../utils/colorUtils';
```
maximusunc, line 19: This import statement should use the @.

src/pages/question/panels/NodePanel.jsx
```diff
@@ -1,5 +1,5 @@
 import React, {
-  useEffect, useRef, useCallback, useContext,
+  useEffect, useRef, useCallback, useContext
```
maximusunc, line 2: Needs a trailing comma.

src/pages/question/panels/NodePanel.jsx
```diff
@@ -13,11 +13,14 @@ import shortid from 'shortid';
...
 
-import entityNameDisplay from '../../../utils/entityNameDisplay';
-import curieUrls from '../../../utils/curieUrls';
+import entityNameDisplay from '@/utils/entityNameDisplay';
+import curieUrls from '@/utils/curieUrls';
+import biolinkUtils from '@/utils/biolink';
 import getNodeTypeColorMap from '../../../utils/colorUtils';
 import NodeProperties from './NodeProperties';
 
+import BiolinkContext from '@/context/biolink';
```
maximusunc, line 22: This import should come before all relative imports.

src/pages/question/panels/NodePanel.jsx
```diff
@@ -121,6 +126,37 @@ export default function NodePanel({ panelStore }) {
...
+      }),
+    );
+    const conceptsSetified = conceptsFormatted.map((c) => ({
+      ...c,
+      name: setify(c.name),
+      set: true,
+    }));
+
+    // Merge concepts with sets interleaved
+    // so that they show up in the list in the
```
maximusunc, line 150: This line has a trailing space.

src/pages/question/panels/NodePanel.jsx
```diff
@@ -121,6 +126,37 @@ export default function NodePanel({ panelStore }) {
...
+      ...c,
+      name: setify(c.name),
+      set: true,
+    }));
+
+    // Merge concepts with sets interleaved
+    // so that they show up in the list in the 
+    // right order
+    const combinedConcepts = conceptsFormatted.map(
+      (c, i) => [c, conceptsSetified[i]] ).flat();
```
maximusunc, line 153: There's an extra space in this line.

src/pages/question/questionBuilder/QuestionGraphViewContainer.jsx
```diff
@@ -1,9 +1,12 @@
 import React, {
-  useState, useEffect, useMemo, useCallback,
+  useState, useEffect, useMemo, useContext
 } from 'react';
 import { FaSpinner, FaPlusSquare } from 'react-icons/fa';
 import { Modal, Button } from 'react-bootstrap';
+
 import queryGraphUtils from '@/utils/queryGraph';
+import API from '@/API';
+import AlertContext from '@/context/alert';
```
maximusunc, line 9: These imports and useContext are no longer used.

maximusunc (changes requested):

maximusunc (approved):

PR #61, opened 2020-10-23
Fixed deployment to work with new API structure

alongreyber:
- Switched over to a reverse proxy container for production deployment.
- Added check for environment in bundle. The bundle will use a locally running robokache version in development.

### DIFF
--- a/webpack.common.js
+++ b/webpack.common.js
@@ -1,5 +1,6 @@
 const path = require('path');
 const HtmlWebpackPlugin = require('html-webpack-plugin');
+const webpack = require('webpack');
 
 const config = {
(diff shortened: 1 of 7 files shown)


webpack.common.js
```diff
@@ -58,6 +59,9 @@ const config = {
       template: './public/index.html',
       favicon: './public/favicon.ico',
     }),
+    new webpack.DefinePlugin({
+      'process.env': JSON.stringify(process.env),
```
maximusunc, line 63: Is this copying over ALL environmental variables?

maximusunc (approved):

PR #62, opened 2020-10-25
Architecture update

alongreyber:
Expanded Docker architecture to to cover local development. This will be helpful in the future when we are relying on the proxy container to handle ARA communication. This also resolves CORS issues and makes the local development environment more similar to production.

Change overview:
- Moved frontend code into subfolder to keep root of repository clean
- Added separate docker-compose and Dockerfiles for development and production
- Created a management script to automate local development environment startup
- Updated README to reflect changes in workflow


maximusunc (changes requested):
There are three things I'd like to see added:
- Get eslint working again (I think just move it into the frontend folder cause we only have manage.js on the outside?) Make sure we keep eslinting while developing locally.
- Fix the syntax in manage.js. It has a lot of linting errors.
- Add two npm script commands to package.json that will run the node manage.js commands. npm run dev, npm run prod, etc.

frontend/queryDispatcher/robokache.js
```diff
@@ -1,14 +1,12 @@
...
-const host = process.env.ROBOKACHE_HOST || 'lvh.me';
-const port = 8080;
-
-const base_url = `${protocol}://${host}:${port}/api/`;
+const base_url = `http://robokache/api/`;
 
 const baseRoutes = {
   async getDocumentData(doc_id, token) {
+    console.log("Getting doc data");
+    console.log(`${base_url}document/${doc_id}/data`)
```
maximusunc, line 9: Remove console.logs

maximusunc (approved):

PR #67, opened 2020-10-28
Name resolver

### LINKED ISSUE #51: Switch curie lookup over to name resolver
http://robokop.renci.org:2433/docs#/lookup/lookup_curies_lookup_post

alongreyber:
Switched node search to use nameResolver and nodeNormalizer for lookup instead of ranker. Closes #51.


frontend/src/pages/question/panels/NodePanel.jsx
```diff
@@ -151,22 +150,77 @@ export default function NodePanel({ panelStore }) {
...
+      if (r.status === 'error') {
+        nodeNormalizerError = r;
+      }
+    });
+
+    if (nodeNormalizerError === 'error') {
+      displayAlert('error',
+        'Failed to contact node normalizer to search curies. You will still be able to select generic types. Please try again later');
+      node.setLoading(false);
+      return;
```
maximusunc, line 211: This seems like it won't work. We're setting nodeNormalizerError to the whole response object if its status is 'error', then checking if that object is equal to the string 'error'. It won't. The forEach can also be a find() so we stop as soon as we get an error. Cause on line 211, we stop and return if there are any errors. Are there any chances one call to the normalizer will fail while others work? Do we want to handle that case, or just show an error if any of them fail?

frontend/src/pages/question/panels/NodePanel.jsx
```diff
@@ -151,22 +150,77 @@ export default function NodePanel({ panelStore }) {
...
+      if (newLength < maxUrlLength) {
+        existingArray.push(c);
+      } else {
+        curiesChunked.push([c]);
+      }
+    });
+
+    const normalizationAPICallResponses = await Promise.all(
+      curiesChunked.map((cs) => API.nodeNormalization.getNormalizedNodes(cs)),
+    );
```
maximusunc, line 197: This all looks reasonable. But silly. We should open an issue or something on the Normalizer repo and see if they can accept a POST with body so we don't have to do this.

frontend/src/pages/question/panels/NodePanel.jsx
```diff
@@ -151,22 +150,77 @@ export default function NodePanel({ panelStore }) {
...
+        'Failed to contact node normalizer to search curies. You will still be able to select generic types. Please try again later');
+      node.setLoading(false);
+      return;
     }
+
+    const curiesWithInfo = {};
+    normalizationAPICallResponses.forEach((r) => Object.assign(curiesWithInfo, r));
+
+    node.updateCuries(
+      Object.values(curiesWithInfo).filter((c) => c).map((c) => ({
```
maximusunc, line 218: What is this filter for?

maximusunc (changes requested):
Our routing solutions seem very haphazard. Is there a disadvantage to just passing everything through the proxy so we don't have to worry about CORS?

frontend/src/pages/question/panels/NodePanel.jsx
patrickkwang: The fix is already in the works, but let's go ahead with a workaround in case they're slow.

maximusunc (changes requested):
It looks like you have a missing trailing comma in nameResolver from the eslint check. I would also like a reply to my overall comment from my first review.

alongreyber:
> Our routing solutions seem very haphazard. Is there a disadvantage to just passing everything through the proxy so we don't have to worry about CORS?

The disadvantages I can think of:
- Possibly worse performance for the client because there may be a bottleneck with the proxy if we send out a lot of requests at once
- Higher load on the server

These disadvantages are definitely small. Switching to redirecting everything through the proxy would save us some complexity, but not a significant amount because of the way the code is written. I think the advantages are small as well. Overall, definitely a toss-up, but I would lean towards not redirecting everything through the proxy just because we already have it written this way.

patrickkwang:
If it's more or less a neutral change and things are working right now, I recommend leaving it alone until/unless it feels more urgent in the future.

proxy/nginx.conf
```diff
@@ -24,6 +24,10 @@ http {
 		location /api/robokache/ {
 			proxy_pass http://robokache/api/;
 		}
+
+		location /api/external/nodeNormalization/ {
+			proxy_pass https://nodenormalization-sri.renci.org/;
```
maximusunc, line 29: This should probably be a variable, like we did with the frontend url, so it doesn't crash if this host isn't up yet.

maximusunc (changes requested):
I'm not sure if this is just my janky docker situation, but I'm getting 404s from the normalizer for everything. Or maybe the service is down right now.

I don't know if we want to do anything, but when the proxy service logs these routes out in the running console, it includes the whole route, which is massive since we're including every curie in the actual route, so the log is super swamped.

maximusunc (approved):

PR #69, opened 2020-10-29
Wrap downloaded answers in a query_graph

### LINKED ISSUE #64: Downloaded questions should be wrapped in a query graph
Right now, all we get are a list of nodes and edges. I think we should wrap these in a query graph object to make it clear what these are.

alongreyber:
Closes #64. Downloading questions now wraps the nodes and edges in a query_graph key. The implementation had the side effect of converting all messages stored in robokache to valid reasoner API messages (probably a good thing overall).

Also added whitespace formatting to downloaded questions and answers. This makes it easier to read downloaded questions and answers.


frontend/src/pages/SimpleViewer.jsx
```diff
@@ -39,7 +39,7 @@ export default function SimpleViewer(props) {
 
     const questionId = response.id;
     // Upload question data
-    const questionData = JSON.stringify({ query_graph: messageStore.message.query_graph });
+    const questionData = JSON.stringify({ query_graph: messageStore.message.query_graph }, null, 2);
```
maximusunc, line 42: It doesn't make much sense to me why we'd be putting spaces in any messages sent to robokache, seems like that would just make the files on disk larger. Wouldn't it make more sense to only add spaces when a user wants to download the json for a message?

frontend/src/components/shared/answersetView/StoredAnswersetViewer.jsx
```diff
@@ -35,8 +35,12 @@ export default function StoredAnswersetView({ question_id, answer_id }) {
       return;
     }
 
+    const questionResponseJSON = JSON.parse(questionResponse);
+    const answerResponseJSON = JSON.parse(answerResponse);
```
maximusunc, line 39: I'm not sure I understand here. Why were these parses added, and why weren't we doing this before?

maximusunc (changes requested):

frontend/src/components/shared/answersetView/StoredAnswersetViewer.jsx
maximusunc: Ok cool. This sounds good.

maximusunc (approved):

PR #70, opened 2020-10-29
Created github workflow for running eslint

alongreyber:


patrickkwang:
Any way to fix the bot's formatting?

patrickkwang:
I think this will be most useful if we can configure the linter precisely so that we will not accept PRs with any errors and/or warnings.

alongreyber:
I'm working on fixing the formatting. Only way to test this is to actually run a PR so that's why this is open now. I can configure this to reject if there are errors, but right now there are a ton from old files so that would stop us from doing any PRs until those are all resolved.

patrickkwang:
Oh it's not so bad. One big run-through would pick up most of these. Half of them are prefer-const errors, which I think I know how to fix.

PR #80, opened 2020-11-04
Tests

maximusunc:
I'm porting over changes from the testing-infrastructure branch because the git trail got off a little.


alongreyber (commented):
Glad you were able to get this working! Thanks for the help. Not sure if this is something we want to include in a future PR or this one - but it would be nice to have coverage printed as well. [There's a Github Action we can use](https://github.com/marketplace/actions/jest-annotations-coverage) instead of add-pr-comment to add Jest code annotations for errors as well as coverage reports.

docker-compose.test.yml
```diff
@@ -0,0 +1,7 @@
+version: '3.5'
+
+services:
+  frontend:
+    build:
+      dockerfile: Dockerfile.test
+    command: npm run test:cov
```
maximusunc, line 7: This command doesn't exist anymore

frontend/.gitignore
```diff
@@ -1,2 +1,4 @@
 node_modules
 pack
+coverage
+jest.results.json
```
maximusunc, line 4: This was just for an action we aren't using, so can be gotten rid of. I'd like to keep coverage here though if we ever run tests purely locally.

.github/workflows/dev.yml
```diff
@@ -37,3 +37,41 @@ jobs:
...
+
+      - name: Copy coverage output from container
+        run: |
+          docker cp frontend:/app/coverage/lcov.info .
+
+      - name: Report coverage results
+        uses: mszlgr/lcov-reporter-action@v0.2.23
+        with:
+          github-token: ${{ secrets.GITHUB_TOKEN }}
+          lcov-file: ./lcov.info
```
maximusunc, line 77: I'm confused by why this is working. On line 58 we're running npm run test, which runs the tests, but doesn't give coverage. Where is this lcov.info coming from?

maximusunc: Oh I see it in the jest.config now. I don't want to do that. I want to be able to run tests without running the whole coverage.

alongreyber (approved):

PR #82, opened 2020-11-05
Template implementation

alongreyber:
Re-implemented template functionality.


frontend/src/questionTemplates/cop_disease.json-e
```diff
@@ -0,0 +1,80 @@
+{
```
maximusunc, line 1: What are all of these json-e files?

frontend/src/pages/question/questionBuilder/questionTemplate/QuestionTemplate.jsx
```diff
@@ -44,19 +48,24 @@ function displayQuestion(questionName) {
...
   const [labels, setLabels] = useState([]);
   const [curies, setCuries] = useState([]);
-  const curieSelector = useRef({});
 
-  function setFocus(num) {
-    curieSelector.current[`curie${num}`].curieSelector.input.focus();
+  // Using a ref here ensures that we have an up-to-date nameList value
+  // in the setFocus method
+  const nameListRef = useRef(null);
+  nameListRef.current = nameList;
```
maximusunc, line 63: You can initialize a ref with a default value on the line above instead of giving it null.

maximusunc (commented):
I'm not done with this review because my head is starting to hurt. Could you explain your reasoning behind adding the CurieConceptSelector and FillIdentifier? This thing was already really gnarly, and I feel like adding more intermediate components and passing refs everywhere is making it even more unsustainable.

alongreyber:
> I'm not done with this review because my head is starting to hurt. Could you explain your reasoning behind adding the CurieConceptSelector and FillIdentifier? This thing was already really gnarly, and I feel like adding more intermediate components and passing refs everywhere is making it even more unsustainable.

frontend/src/questionTemplates/cop_disease.json-e
alongreyber: Not sure. They were in the old robokop repo, so I copied them over, but I never looked at them.

maximusunc: I'm not seeing them in the old repo. Please remove them.

maximusunc:
I would like to revert back to the way it was structured before. I decided to duplicate the curie selector logic before because it kept the ref logic simpler. Please just use the existing code, but update it to use hooks and such.

alongreyber:
> I would like to revert back to the way it was structured before. I decided to duplicate the curie selector logic before because it kept the ref logic simpler. Please just use the existing code, but update it to use hooks and such.

Do you mind if I take one more try at simplifying the ref logic? There is more logic in the curie selector than there was when it was duplicated (include the API calls to nodeNormalizer and nameResolver), so I think there's more of a benefit to sharing a component than there was in the original architecture. I would rather try to fix the new structure before reverting.

maximusunc:
> > I would like to revert back to the way it was structured before. I decided to duplicate the curie selector logic before because it kept the ref logic simpler. Please just use the existing code, but update it to use hooks and such.
> 
> Do you mind if I take one more try at simplifying the ref logic? There is more logic in the curie selector than there was when it was duplicated (include the API calls to nodeNormalizer and nameResolver), so I think there's more of a benefit to sharing a component than there was in the original architecture. I would rather try to fix the new structure before reverting.

Yeah, you can keep working on it if you want. My priorities right now are that we just get the template working again while not making it any more complex and come back to it later for improvements (your implementation works, but it would be more complex to dive into later). If you think you can also improve it in a reasonable amount of time, then by all means, but if not then we just need to get it working and move on.

frontend/src/components/shared/curies/CurieConceptSelector.jsx
```diff
@@ -0,0 +1,189 @@
...
+}) {
+  // Reference to the input element used for setting focus
+  const inputRef = useRef(null);
+  useEffect(() => {
+    if (!focus) return;
+    // We have to use the deprecated find-dom-node method
+    // because react-bootstrap doesn't expose an alternative interface
+
+    // eslint-disable-next-line react/no-find-dom-node
+    const node = ReactDOM.findDOMNode(inputRef.current);
```
maximusunc, line 33: The property on the FormControl should be inputRef instead of just ref and will set the correct ref so you can just call inputRef.current.focus().

frontend/src/components/shared/curies/CurieConceptSelector.jsx
```diff
@@ -0,0 +1,189 @@
...
+    node.focus();
+  }, [focus]);
+
+  // After we have set focus, clear the variable so we can set it again
+  useEffect(() => { if (focus && clearFocus) clearFocus(); }, [focus]);
```
maximusunc, line 38: This logic can go in the above useEffect. Also, I don't think we should need this logic. We should be able to simply call focus() on an input and forget about it, not go and reset some property.

maximusunc (changes requested):

frontend/src/components/shared/curies/CurieConceptSelector.jsx
alongreyber: This logic is necessary if we want to be able to set focus again after setting it once. If we set focus to true again, it won't call the useEffect without this logic.

maximusunc: I think what I'm getting at is that now we aren't passing refs, which is good, but now we have this other random property that we have to pass around. The reason I initially decided to duplicate the curie container logic was for this reason, that all the logic would be contained in one component and simpler to follow.

patrickkwang:
I am testing the manually and ran into a problem. To reproduce:

1. Ask a Quick Question
1. Template
1. "Find a drug to treat..."
1. "ebola" -> MONDO:0005737
1. Load Question
1. Submit

This receives a 422 error from messenger. I'm seeing the same behavior for other template questions as well, but not for simple custom questions.

Also of note: clicking Ask a Quick Question from the error page does nothing.

frontend/src/components/shared/curies/CurieConceptSelector.jsx
alongreyber: I think that the decision to duplicate the curie container logic made sense when the curie container was simple and only pulled the curies from a JSON file. However, the number of lines (and complexity) in the NodePanel and useNode files has nearly doubled since then, so I think it now makes sense to try to re-use that logic as best we can.

alongreyber:
> I am testing this manually and ran into a problem. To reproduce:
> 
>     1. Ask a Quick Question
> 
>     2. Template
> 
>     3. "Find a drug to treat..."
> 
>     4. "ebola" -> MONDO:0005737
> 
>     5. Load Question
> 
>     6. Submit
> 
> 
> This receives a 422 error from messenger. I'm seeing the same behavior for other template questions as well, but not for simple custom questions.
> 
> Also of note: clicking Ask a Quick Question from the error page does nothing.

Should be fixed now.

maximusunc:
I think I figured out the alarms going off in my head. You have replaced the BionamesBrowser, CurieSelector, and CurieSelectorContainer components. The only thing that now uses them is the Neighborhood page, which is currently broken and should be fixed (but leave it for now, maybe just hide the button on the landing page). The scope of this PR is expanding and in a perfect world, should be multiple PRs. But we need to get this done before merging back into the old repo. If those components have truly been replaced, then we should delete those files. And please make sure that you have copied over all logic from those components and we haven't lost any functionality.

maximusunc (approved):

PR #84, opened 2020-11-09
Re-implemented QuestionNew Page

### LINKED ISSUE #72: Reimplement Signed In Question Builder page

alongreyber:
Re-implementation of QuestionNew page. Closes #72.


maximusunc:
Why is the gitbot so far behind?

frontend/src/pages/question/QuestionNew.jsx
```diff
@@ -0,0 +1,143 @@
...
+    const data = { query_graph };
+
+    // Transform the data into a json blob and give it a url
+    const json = JSON.stringify(data, null, 2);
+    const blob = new Blob([json], { type: 'application/json' });
+    const url = URL.createObjectURL(blob);
+
+    // Create a link with that URL and click it.
+    const a = document.createElement('a');
+    a.download = 'robokopMachineQuestion.json';
```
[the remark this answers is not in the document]
maximusunc: And this download is exactly the same as the one in simple question. So we should move it down into the question builder component.

frontend/src/pages/question/QuestionNew.jsx
```diff
@@ -0,0 +1,143 @@
...
+    a.href = url;
+    a.click();
+    a.remove();
+  }
+
+  function onResetQuestion() {
+    questionStore.resetQuestion();
```
maximusunc, line 50: We should confirm with the user before just resetting the question builder.

frontend/src/pages/question/QuestionNew.jsx
```diff
@@ -0,0 +1,143 @@
...
+    }
+    const questionMeta = response;
+    questionMeta.metadata.hasAnswers = true;
+    response = await API.cache.updateQuestion(questionMeta, user.id_token);
+    if (response.status === 'error') {
+      displayAlert('error', response.message);
+      return;
+    }
+
+    displayAlert(
```
maximusunc, line 74: This seems like bad UX. We shouldn't make the user click this alert when the answer comes back. And we shouldn't show errors on the questionAnswerViewer page while we wait for an answer. I think when a user submits a question, we can go to the questionAnswerViewer page and show the alert saying we're getting an answer and then just show a loading icon. Once we get the answer, we should automatically show it.

maximusunc (changes requested):

frontend/src/pages/question/QuestionNew.jsx
```diff
@@ -114,30 +88,34 @@ export default function QuestionNew() {
...
-    history.push(`/question/${questionId}`);
-    displayAlert('info', "Fetching answer, we will let you know when it's ready.");
+    pageStatus.setLoading('Fetching answer, this may take a while');
 
     // Start the process of getting an answer and display to user when done
-    getAnswerAndDisplayAlert(questionId);
+    await fetchAnswer(questionId);
+
+    // Redirect to created question
+    history.push(`/question/${questionId}`);
```
maximusunc, line 97: The answer id is returned from API.queryDispatcher.getAnswer. We should move this to the end of fetchAnswer() and include the answer id in the route. Just to make sure we go to the correct answer.

maximusunc: If we wanted to be fancy, once we get the answer back, we could check if the user is still on this page (maybe by looking at location). If they're still here, send them to the answer page, if they're somewhere else, just show the success alert.

maximusunc (changes requested):

maximusunc (approved):

Issue #83, opened 2020-11-09
Fix edge cases with question builder


maximusunc:
- [x] Edge panel header should be colored according to the source and target node types
- [x] switching target and source nodes in editing edge panel, Save button is greyed out
- [x] edge panel, source and target nodes should have their ids in the displayed name. Helps when you have multiple nodes of the same name/type
- [x] edge panel, target node dropdown can have total number of predicates available.
- [x] graph display, each node should have its id in its display name.
- [x] graph display, after adding a second node, the graph doesn't update correctly. i.e. I can't see the second node until I zoom out or otherwise manually cause a page rerender.
- [x] edge panel, source and target should be auto-filled
- [x] edge panel, predicates look like they might not be getting cleared out when panel is closed.
- [x] graph display, something is weird with height, if you add three floating nodes, they start going outside of the display area, but this doesn't happen in the deployed version. might be another getDisplayOptions bug.
- [x] if you have two nodes and an edge, then delete both nodes, the trimFloatingNodes needs to delete them both and the edge.
- [x] reimplement node sets
- [x] nodes with curies should have a bolder border

maximusunc:
With the services we use now, we aren't getting a total number of predicates for the nodes in an edge panel.

maximusunc:
The QuestionGraphView component has some rerendering issues. The missing added node until resizing and the graph height issues are both caused by the graph not rerendering properly. In the mean time, I've uncommented the random key so it rerenders when we need it to, but this also causes more than necessary renders. Issue for another day.

maximusunc:
I was wrong in that an edge can still exist with two deleted nodes (this seems weird, maybe we can change this). But the page crashes when you go to delete an edge when at least one of its nodes is also deleted. So I'm going to fix that.

maximusunc:
I think node sets are already implemented. Hopefully we just need to port back over to gamma viewer.

PR #88, opened 2020-11-11
Question builder update

maximusunc:
Fixes issues mentioned in #83

### DIFF
--- a/frontend/src/components/shared/graphs/QuestionGraphView.jsx
+++ b/frontend/src/components/shared/graphs/QuestionGraphView.jsx
@@ -2,6 +2,7 @@ import React, {
   useEffect, useRef, useMemo,
 } from 'react';
 import _ from 'lodash';
+import shortid from 'shortid';
 
 import getNodeTypeColorMap from '../../../utils/colorUtils';
 import entityNameDisplay from '../../../utils/entityNameDisplay';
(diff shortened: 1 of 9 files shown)


frontend/src/components/shared/graphs/QuestionGraphView.jsx
```diff
@@ -25,7 +26,7 @@ function defaultNodePreProc(n) {
       y: -10,
     };
   }
-  if (n.curieEnabled) {
+  if ((Array.isArray(n.curie) && n.curie.length) || (n.curie && typeof n.curie === 'string')) {
```
alongreyber, line 29: Curie should never be a string here. When we upload it the string is converted to an array. I tested it by uploading a query_graph with a string curie.

frontend/src/pages/question/panels/EdgePanel.jsx
```diff
@@ -67,7 +67,7 @@ export default function EdgePanel(props) {
     );
   }
 
-  const filteredPredicateList = getFilteredPredicateList() || [];
+  const filteredPredicateList = useMemo(getFilteredPredicateList, [edge.sourceId, edge.targetId]) || [];
```
alongreyber, line 70: I think this needs predicateList and biolink as dependencies

alongreyber (changes requested):

alongreyber (approved):

PR #105, opened 2020-11-12
Re-implement question forking

alongreyber:


maximusunc (approved):
My only comment is that we get the question list every time we open the modal. Yes, the question list could change while we're on this page, but I'm not concerned with that. I feel like we could easily just toggle the the modal, and if it's open and we don't already have questions, then go grab them. But as long as the question list api call is really snappy, maybe we don't  have to worry about getting it every time (or really, a user isn't going to be reopening this thing usually.)

PR #110, opened 2020-11-20
Set maxContentLength and maxBodyLength for all Axios requests

alongreyber:
Used a utility to return a base configuration for all axios requests that includes maxContentLength and maxBodyLength. Updated both queryDispatcher and frontend API files.

### DIFF
--- a/frontend/queryDispatcher/index.js
+++ b/frontend/queryDispatcher/index.js
@@ -1,5 +1,10 @@
+const axios = require('axios');
 const router = require('express').Router();
 const routes = require('./routes');
+
+// Set axios request max size
+axios.defaults.maxContentLength = Infinity;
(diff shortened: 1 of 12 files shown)


maximusunc (dismissed):
1. I would make the baseConfig just an object in the utils instead of a function. you're always spreading it anyways which makes a copy.
2. I would make all imports consistent, I lean towards { baseConfig, handleAxiosError }
3. You can remove the let config = baseConfig() and just instantiate config like:
`const config = { ...baseConfig, etc. }`

alongreyber:
Would you like me to merge this or make the changes you requested?

maximusunc:
Please make the changes. Shouldn't take long.

maximusunc (approved):

PR #111, opened 2020-11-30
Switched to Github Container Registry

alongreyber:
Switched to Github Container Registry and added CI for pushing both proxy and frontend containers on commits to main.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -8,7 +8,7 @@ on:
 
 jobs:
     push_frontend:
-        name: Push Docker image to GitHub Container Registry.
+        name: Push frontend image to GitHub Container Registry.
         runs-on: ubuntu-latest
         steps:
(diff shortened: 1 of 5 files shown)


.github/workflows/main.yml
```diff
@@ -0,0 +1,38 @@
+name: 'Push to Github Container Registry'
+
+on:
+    pull_request:
+        types: closed
```
maximusunc, line 5: Is there a types: merged? I don't think it's a huge deal, but this would trigger on non-merged closed PRs as well.

maximusunc (approved):

.github/workflows/main.yml
alongreyber: That's a good point. Surprisingly, Github Actions doesn't offer a `types: merged`. It's still possible, but a bit more work. I'll open an issue for that separately.

PR #124, opened 2020-12-04
Create build artifacts on PRs

alongreyber:
Create build artifacts of the following:
- Test results
- Coverage
- ESLint errors

All artifacts are formatted as HTML pages.
Closes #124


maximusunc:
I don't think I want to go this route. I don't think anyone will actually go and see these artifacts if it requires going to the action, downloading a zip file, and then opening that index.html in the browser. What I'm envisioning (and really actually isn't all that *necessary*, so doesn't need to be high priority) is that the test output is included in the action log.

Ex: 
Screen Shot 2020-12-04 at 11 04 47 AM

maximusunc:
I'm going to close this. We can revisit other solutions at a later time.

PR #127, opened 2020-12-09
Fix bug with CurieConceptSelector

alongreyber:
To reproduce issue:

1. Visit http://robokop.renci.org:7080/simple/question
1. Blank question
1. New node
1. Search for "Coronavirus"
1. Click select
1. Nothing happens

This is caused by the code not considering a selection valid if it does not have a type. This pull request updates the definition of a valid type to include a curie.

### DIFF
--- a/frontend/src/components/shared/curies/CurieConceptSelector.jsx
+++ b/frontend/src/components/shared/curies/CurieConceptSelector.jsx
@@ -116,7 +116,10 @@ export default function CurieConceptSelector({
   }
 
   const nRows = concepts.length + curies.length;
-  const showOptions = searchTerm && !selection.type;
+
+  // Show options if there is no curie or type (valid selection) AND we have a search term
+  const showOptions = searchTerm && !selection.type && !selection.curie.length;
+
   const isEmpty = nRows === 0;
 
   const rowHeight = 50;
@@ -134,7 +137,7 @@ export default function CurieConceptSelector({
             inputRef={(ref) => { inputRef.current = ref; }}
             onChange={(e) => updateSearchTerm(e.target.value)}
           />
-          {!showOptions && selection.curie.length > 0 && (
+          {!!selection.curie.length && (
             <InputGroup.Addon>
               {selection.curie[0]}
             </InputGroup.Addon>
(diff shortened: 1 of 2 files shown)


maximusunc:
This is a much farther reaching issue that just the curie selector. Just search for "node.type" across the whole repo. With old Robokop and its KP, every node had a type and therefore we have a bunch of logic assuming type. This PR is just fixing one instance where we want to not care about type, but it's not that simple. Do we even know if Messenger currently accepts nodes without a type?

maximusunc:
On this branch, when I follow the instructions listed above and then ask the question with just the one node, I get no answers back. When I go and download the returned answerset, it's just an empty object with no query graph, knowledge graph, or results. So something went wrong somewhere, and I don't see anything logged anywhere.

maximusunc:
This is going to open a can of worms, and not high priority. Let's leave this open for reference later, but not doing anything with it for now.

alongreyber:
It seems like the bug itself is high priority - we can't build questions that have curies in them. Is there a different route that you want me to work on for fixing this?

maximusunc:
#89 would probably be the way to go. I can try and track down the code from the old repo for this.

maximusunc:
We did something simple like this: https://github.com/NCATS-Gamma/robokop/blob/master/src/AppConfig.jsx#L503 where someone could input a plain curie in the field and we'd accept it. In this new case, we can probably just grab the curie from the selected option. But then we still get to the type missing issue. I still need to think through how we want to handle this. So keep this on hold for now.

maximusunc:
The link in the last comment is just one part of what we were doing before. https://github.com/NCATS-Gamma/robokop/blob/master/src/stores/newQuestionStore.js#L479 shows the validation in the node panel. So we were handling no node type, but we also had validation on the curie that we should re-add. So I think we can continue with this fix, but with the extra validation.

- [ ] A node can have no type, but then it has to have a valid curie. This also includes a user being able to input a curie directly and we don't need to do any of the node lookup. This is advanced user, so it's up to them to type in a real curie.

I just opened a new issue #136 for this.

PR #142, opened 2020-12-14
Convert biolink node types to normalized types

### LINKED ISSUE #137: Handle dynamic node types from Node Normalizer
If I search for Ebola, it returns with a type of "biolink:Disease". In the useNode.js hook, we're still using the concepts listed in the config.json we have, so the type we get back isn't found and we basically treat it as not having a type. We should probably just take whatever type we get from biolink, and figure out a smarter way to choose colors for the UI when displaying nodes.

### LINKED ISSUE #141: Convert Node Normalizer call to POST
There's a POST method that got added recently - might try using that if you are not already

_Originally posted by @cbizon in https://github.com/NCATS-Gamma/robokop-rewrite/issues/140#issuecomment-744593217_

maximusunc:
Fixes #141 and fixes #137

### DIFF
--- a/frontend/src/pages/question/panels/NodePanel.jsx
+++ b/frontend/src/pages/question/panels/NodePanel.jsx
@@ -17,6 +17,22 @@ import biolinkUtils from '@/utils/biolink';
 
 import CurieConceptSelector from '@/components/shared/curies/CurieConceptSelector';
 
+function normalizeNodeType(type) {
+  let normalizedType = type;
+  if (type.indexOf(':') > -1) {
+    [, normalizedType] = type.split(':'); // grab the second item, the type
+    const splitRegex = new RegExp(/(?<=[a-z])[A-Z]|[A-Z](?=[a-z])/g);
+    const splitByCapital = normalizedType.replaceAll(splitRegex, (match, ind) => {
+      if (ind !== 0) {
+        return `_${match.toLowerCase()}`;
+      }
+      return match.toLowerCase();
+    });
+    normalizedType = splitByCapital;
+  }
+  return normalizedType;
+}
+
 /**
  * Node Panel
  * @param {Object} panelStore panel custom hook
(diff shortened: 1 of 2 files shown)


frontend/src/pages/question/panels/NodePanel.jsx
```diff
@@ -17,6 +17,22 @@ import biolinkUtils from '@/utils/biolink';
 
 import CurieConceptSelector from '@/components/shared/curies/CurieConceptSelector';
 
+function normalizeNodeType(type) {
```
alongreyber, line 20: I think there's probably a better name than normalize. Especially because biolink is switching to use "biolink:Gene" as the "normal" name, and it's really robokop that uses the "abnormal" formatting.

alongreyber (approved):

frontend/src/pages/question/panels/NodePanel.jsx
maximusunc: I changed the function name and added some documentation. I also looked at the biolink model yaml file and the types haven't changed, my thought is that the node normalizer service is adding the biolink part to show where it was getting its types and we shouldn't be keeping that info. And I also think KGs are not going to have those types but just have 'disease'. Though I don't really understand why the types need this prefix. @patrickkwang do you have any insight?

patrickkwang: My feeling is that the biolink model itself is the only place that should _not_ use the `biolink:` prefix. Think of it like a namespace. KGs, KPs, ARAs, etc. will all use the full CURIE form.

maximusunc: Ok, so I think we should still convert the node types to what we already expect internally. This will allow us to continue to send valid queries to Messenger and receive answers back. Down the road, when we add new ARAs or Messenger gets updated, then we can remove this conversion.

I think that there are other things we can take care of in the mean time to make node types more dynamic, like #146 so that when we remove the conversion, everything will still work.

alongreyber (approved):

PR #147, opened 2020-12-18
Update parse message function

### LINKED ISSUE #40: Update parseMessage and error reporting for ingested messages
I tried the Ask ARA functionality with the covidkop response. No results or knowledge_graph were returned and the error message was "Unable to parse message". I think we should change that error to be more descriptive when there are no answers returned. My thinking is that we parse those messages without error and then display an error message in the AnswersetView.

_Originally posted by @alongreyber in https://github.com/NCATS-Gamma/robokop-rewrite/pull/37#discussion_r501756507_

maximusunc:
Closes #40

### DIFF
--- a/frontend/src/pages/simplecss.css
+++ b/frontend/src/pages/simplecss.css
@@ -49,7 +49,7 @@ a .list-group-item-text {
     cursor: pointer;
 }
 .resetButton {
-    background-color: #f9cbc2 !important;
-    color: darkred !important;
-    margin-left: 20px !important;
+    background-color: #f9cbc2;
+    color: darkred;
+    margin-left: 20px;
 }
--- a/frontend/src/pages/simplecss.css
+++ b/frontend/src/pages/simplecss.css
@@ -48,3 +48,8 @@ a .list-group-item-text {
     margin-bottom: 20px;
     cursor: pointer;
 }
+.resetButton {
+    background-color: #f9cbc2 !important;
+    color: darkred !important;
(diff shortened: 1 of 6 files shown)


frontend/src/utils/parseMessage.js
```diff
@@ -28,45 +28,60 @@ export default function parseMessage(message) {
...
+    // results can be empty
+    if (message.results.length) {
+      if ('node_bindings' in message.results[0]) {
+        if (!Array.isArray(message.results[0].node_bindings)) {
+          // if results isn't in the newer format, we're gonna try and convert it
+          message.answers = message.results;
+          delete message.results;
+        }
+      } else {
+        throw Error('Error parsing message results. Results should have node_bindings and edge_bindings.');
```
alongreyber, line 43: This error mentions edge_bindings but doesn't actually check for an edge_binding property.

frontend/src/pages/simplecss.css
```diff
@@ -47,4 +47,9 @@ a .list-group-item-text {
     border-radius: 4px;
     margin-bottom: 20px;
     cursor: pointer;
+}
+.resetButton {
+    background-color: #f9cbc2 !important;
```
alongreyber, line 52: Usually `!important` isn't good practice. We should probably use one of [these methods](https://material-ui.com/customization/components/) for styling a reset button.

alongreyber (changes requested):

alongreyber (approved):

PR #150, opened 2020-12-22
Convert to use concepts context from biolink model

### LINKED ISSUE #146: useNode hook still uses concepts from the hardcoded config file

maximusunc:
This turned into a bit bigger change than anticipated. I removed the config files because we don't need them, and now concepts are read in from biolink at the beginning and put into a context that any components can use. We use concepts quite extensively throughout the app, so it's possible I missed some places, but my manual testing has all worked fine.

Fixes #146


frontend/src/App.jsx
```diff
@@ -25,26 +25,33 @@ import API from '@/API';
...
     const response = await API.biolink.getModelSpecification();
     if (response.status === 'error') {
       simpleSetAlert('error',
         'Failed to contact server to download biolink model. You will not be able to select predicates. Please try again later');
       return;
     }
+    const validConcepts = Object.keys(biolinkUtils.getValidConcepts(response));
     setBiolink(response);
+    setConcepts(validConcepts);
+    console.log(validConcepts);
```
alongreyber, line 54: Remove console.log

frontend/src/App.jsx
```diff
@@ -25,26 +25,33 @@ import API from '@/API';
...
 import BiolinkContext from '@/context/biolink';
+import ConceptsContext from '@/context/concepts';
+
+import biolinkUtils from '@/utils/biolink';
 
 export default function App() {
   const [user, setUser] = useState(null);
+  const [biolink, setBiolink] = useState(null);
+  const [alert, setAlert] = useState({});
+  const [concepts, setConcepts] = useState([]);
```
alongreyber, line 36: This is duplicated state, where we could just use
```javascript
const concepts = Object.keys(biolinkUtils.getValidConcepts(biolink));
```

I think we should try to never duplicate state if at all possible.

frontend/src/App.jsx
```diff
@@ -54,55 +61,57 @@ export default function App() {
...
-                      </Route>
-                      <Route path="/">
-                        <Landing isSignedIn={isSignedIn} />
-                      </Route>
-                    </Switch>
-                  </div>
-                  <Footer />
-                </StylesProvider>
-              </ThemeProvider>
+              <ConceptsContext.Provider value={concepts}>
```
alongreyber, line 64: Context providers are kinda ugly and I think we should try to minimize their use as much as possible. We could get the same information as the ConceptsContext using the BiolinkContext. I think it's cleaner to just pull that information from the BiolinkContext whenever we need it rather than creating a new context for Concepts because the concepts exist in the biolink context.

This can be done in only a couple lines:
```javascript
const biolink = useContext(BiolinkContext);
const concepts = biolinkUtils.getValidConcepts(biolink);
```

alongreyber (changes requested):

frontend/src/App.jsx
maximusunc: We use concepts in many places, whereas we only use one other property of biolink in the edge panel. I would like to keep the concepts context for these reasons:
- it might be extra in App, but then we'd have to add an extra line in at least 8 different places
- we'd only be importing a small-ish array instead of the whole biolink object
- We do the getValidConcepts operation once to get a static array instead of in a bunch of different components, and then we might need to memoize those so we don't do the operation on every new render.

frontend/src/App.jsx
maximusunc: See my other comment.

alongreyber: Even if we keep the ConceptsContext, I think we should remove the extra state here. We can change the concepts to the following to not recompute on re-renders:

```javascript
const concepts = useMemo( Object.keys(biolinkUtils.getValidConcepts(biolink)), [biolink] );
```

frontend/src/pages/question/panels/NewQuestionPanelModal.jsx
```diff
@@ -14,13 +14,14 @@ import './panels.css';
...
 export default function NewQuestionPanelModal({ panelStore, onQuestionUpdated }) {
+  const concepts = useContext(ConceptsContext);
   /**
    * Get the panel background color
    * @param {Boolean} isNodePanel is panel of type node
    */
   function getBackgroundColor(isNodePanel) {
     // set the color of the node/edge panel header
-    const panelColorMap = getNodeTypeColorMap(config.concepts);
+    const panelColorMap = getNodeTypeColorMap(concepts);
```
alongreyber, line 24: These shouldn't be gray right?
Screen Shot 2020-12-23 at 9 30 17 AM

maximusunc: They should. We have a hard-coded index of types to colors, and any type not in that index will be grey. If we want those to have colors, we just need to add them to the index and give them colors, but that should be dealt with in a separate PR.

frontend/src/App.jsx
maximusunc: We could do it this way, but we'd also have to add some extra code to the getValidConcepts function to check if biolink is null (because it will be on the initial render). I'm really ok with the extra state how it is. It's not a complete duplicate of state, as we're converting the biolink classes to a trimmed down list.

alongreyber (approved):
For the record - I still don't like the duplication of state in this PR. Duplicated state like this can create issues where state is out of sync and seems like bad practice. Especially because this can be fixed by changing only one line.

maximusunc:
I understand your reservations. However, we're only GETting the biolink model on initial render, so it won't be changing and getting out of sync. Also, it wouldn't be a one line change because if we memoized concepts, we would then need to check that biolink wasn't null before getting valid concepts.

PR #152, opened 2020-12-22
Allow manual curie in node panel

### LINKED ISSUE #89: Fix defining nodes by a curie

maximusunc:
Does basically everything in #127 but also allows manual curies in the node panel.

Closes #89


alongreyber (changes requested):
Looks good, but need to fix merge conflicts with main.

alongreyber (changes requested):
This doesn't seem to work. Typing in a curie (anything with a colon) displays loading but makes no API calls and seems to sit there forever in an invalid state. I've verified this doesn't happen before the most recent commit:


https://user-images.githubusercontent.com/1634866/103662073-818c7f80-4f3d-11eb-94fa-eb1310a610ef.mov

alongreyber (approved):

PR #186, opened 2021-05-25
Filter out webpack hot reload logs from nginx

maximusunc:


proxy/nginx.conf
```diff
@@ -48,9 +56,6 @@ http {
 		location ~ /api/external/nodeNormalization/(.*) {
 			set $url https://nodenormalization-sri.renci.org/;
 			proxy_pass $url$1$is_args$args;
-
-			# Hide long parameters in log
-			access_log /dev/stdout combined_no_query;
```
alongreyber, line 53: If you remove this line you should remove the definition of combined_no_query at the top of this file.

alongreyber (changes requested):

proxy/nginx.conf
maximusunc: I'm using it on line 27 because we still want to hide long parameters.

alongreyber (approved):

PR #187, opened 2021-05-27
New answer page

maximusunc:


frontend/src/utils/d3Utils.js
```diff
@@ -0,0 +1,180 @@
+import * as d3 from 'd3';
```
alongreyber, line 1: We have utils repeated in this file name. If there's no way to categorize these utils beyond "d3" maybe they could live in the index file.

alongreyber (changes requested):

frontend/src/pages/answer/kgBubble/KgBubble.jsx
```diff
@@ -0,0 +1,111 @@
+/* eslint-disable indent, no-use-before-define, func-names, no-return-assign */
```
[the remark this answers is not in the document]
maximusunc: Yeah, it's not great, but I'm aware of the behavior.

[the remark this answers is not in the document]
maximusunc: I don't think this would work out, because a node bubble can be referenced in multiple results. But I think it could filter the results and only show the ones that contain that node. Created #188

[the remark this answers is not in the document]
maximusunc: This is an interesting idea. I wonder how it would work with the physics. Would you expect a hovered node to push the surrounding nodes away when it expands, or overlap? And then would it shrink back down if you started to drag it? Opened #189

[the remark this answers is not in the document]
maximusunc: Yeah, we definitely need to handle this differently. A slider is one idea, but a user could still allow all nodes and get back to this state. The idea I had with the full knowledge graph was to make it a non-interactive canvas when the number of nodes went above a certain threshold. Opened #190

frontend/src/pages/answer/Answer.jsx
```diff
@@ -0,0 +1,124 @@
+import React, { useContext } from 'react';
```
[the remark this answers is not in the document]
maximusunc: This message is interesting. It has a query graph and a full knowledge graph, but the results are empty. Makes sense that the results table doesn't show up. The kg bubble graph doesn't show up because it relies on the results to get node bubble sizes and therefore isn't showing up either (This seems expected behavior to me). I can add a message in the UI saying that there are no results for this message so it's clear the UI didn't "break".

frontend/src/pages/answer/resultsTable/SupportingPublications.jsx
```diff
@@ -0,0 +1,80 @@
+import React, { useReducer, useEffect } from 'react';
```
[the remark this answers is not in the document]
maximusunc: @sarah-seitanakis do you have any suggestions? I'm not sure what would go there in its place, or whether we make the result explorer graph bigger (though I don't think it needs extra space).

frontend/src/pages/answer/kgBubble/KgBubble.jsx
```diff
@@ -0,0 +1,111 @@
...
+import React, {
```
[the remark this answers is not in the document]
maximusunc: I'm controlling the nodes to not go outside the bounds of the box. Though I'm not sure letting the bubbles go just slightly outside the box is what we want. @sarah-seitanakis do you have any input?

frontend/src/pages/answer/resultsTable/ResultExplorer.jsx
```diff
@@ -0,0 +1,261 @@
+/* eslint-disable indent, no-use-before-define, func-names, no-return-assign */
```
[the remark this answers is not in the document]
maximusunc: Part of the idea here was so that the user could see which nodes changed between different results. For a result with two nodes, it's harder to see, but for a longer query graph and two results that just have one node different, it'd make more of an impact.

frontend/src/pages/answer/Answer.jsx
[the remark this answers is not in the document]
maximusunc: Opened #192. This would definitely be nice to have. The graphs grab their size on initial render, so if you resize and then refresh the screen, they'll fit again.

frontend/src/utils/d3Utils.js
maximusunc: Deleted this extra file in b7ffb6d03a940b75d2f88f941ff8b25e8effee72

frontend/src/pages/answer/Answer.jsx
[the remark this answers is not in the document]
maximusunc: Added no results message in 28e84e975b3a613682d903cba2e1558f156e689f

frontend/src/pages/answer/resultsTable/ResultsTable.jsx
```diff
@@ -0,0 +1,119 @@
+import React, { useMemo } from 'react';
```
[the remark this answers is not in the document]
maximusunc: Fixed in 542aeee587aae9ba823e47d66ed21bb03d68a851

frontend/src/pages/answer/fullKg/KgFull.jsx
```diff
@@ -0,0 +1,127 @@
...
+    edges.forEach(drawEdge);
+    context.strokeStyle = '#aaa';
+    context.stroke();
+
+    nodes.forEach(drawNode);
+
+    context.restore();
+  }
+
+  useEffect(() => {
```
patrickkwang, line 69: I think it might help readability if big useEffect functions were not inlined, and were named.

frontend/src/pages/answer/fullKg/KgFull.jsx
```diff
@@ -0,0 +1,127 @@
...
+    d3.select(canvasRef.current)
+      .attr('width', 0)
+      .attr('height', 0);
+    if (Object.keys(message).length) {
+      const simulationWorker = new Worker();
+      const kgLists = kgUtils.getFullDisplay(message);
+      toggleLoading(true);
+      simulationWorker.postMessage(kgLists);
+
+      simulationWorker.onmessage = (e) => {
```
patrickkwang, line 79: It makes me nervous that we appear to be posting a message to this Worker _before_ setting its `onmessage` handler.

frontend/src/pages/answer/fullKg/__mocks__/simulation.worker.js
```diff
@@ -0,0 +1,12 @@
...
+    this.url = stringUrl;
+    this.onmessage = () => {};
+  }
+
+  postMessage(msg) {
+    this.onmessage(msg);
+  }
+}
+
+export default Worker;
```
patrickkwang, line 12: I suspect that I could use some documentation here. Why do we need this? Does posting a message actually do nothing?

frontend/src/pages/answer/fullKg/simulation.worker.js
```diff
@@ -0,0 +1,21 @@
...
+import * as d3 from 'd3';
+
+onmessage = (e) => {
+  const { edges, nodes } = e.data;
+  const simulation = d3.forceSimulation(nodes)
+    .force('charge', d3.forceManyBody())
+    .force('collide', d3.forceCollide(10).strength(0.5))
+    .force('link', d3.forceLink(edges).id((d) => d.id).distance(50).strength(1))
+    .force('forceX', d3.forceX())
+    .force('forceY', d3.forceY())
```
patrickkwang, line 11: Could we document what each of these forces is meant to do?

frontend/src/pages/answer/fullKg/simulation.worker.js
```diff
@@ -0,0 +1,21 @@
+/** Web Worker to run d3 force simulation and return placed nodes and edges */
```
patrickkwang, line 1: How does this work? In KgFull.jsx we import Worker from `./simulation.worker`, but there is no Worker defined here as far as I can tell.

frontend/src/pages/answer/kgBubble/KgBubble.jsx
```diff
@@ -0,0 +1,111 @@
...
+    }
+  }, [nodes]);
+
+  function drawBubbleGraph() {
+    const svg = d3.select(svgRef.current);
+    const { width, height } = svg.node().parentNode.getBoundingClientRect();
+    // clear the graph
+    svg.selectAll('*').remove();
+    const getNodeRadius = kgUtils.getNodeRadius(width, height, numQgNodes, numResults);
+    const converted_nodes = nodes.map((d) => ({ ...d, x: Math.random() * width, y: Math.random() * height }));
```
patrickkwang, line 50: I wonder if it would make the subsequent simulation easier if we start with quasi-random positions rather than uniformly random ones. Something like a Halton sequence: https://en.wikipedia.org/wiki/Halton_sequence

This would also mean that the positions are in fact deterministic, which could be nice.

frontend/src/pages/answer/kgBubble/KgBubble.jsx
```diff
@@ -0,0 +1,111 @@
...
+              return name || id || 'Something';
+            })
+            .each(graphUtils.ellipsisOverflow));
+
+    simulation.alpha(1).restart();
+  }
+
+  useEffect(() => {
+    if (nodes.length) {
+      drawBubbleGraph();
```
patrickkwang, line 97: I like this arrangement, where we separate the _idea_ of drawing the bubble graph from _when_ we do it (when the nodes change).

frontend/src/pages/answer/resultsTable/SupportingPublications.jsx
```diff
@@ -0,0 +1,80 @@
...
+import Paper from '@material-ui/core/Paper';
+import List from '@material-ui/core/List';
+import ListItem from '@material-ui/core/ListItem';
+import ListItemText from '@material-ui/core/ListItemText';
+import Collapse from '@material-ui/core/Collapse';
+import ExpandLess from '@material-ui/icons/ExpandLess';
+import ExpandMore from '@material-ui/icons/ExpandMore';
+import shortid from 'shortid';
+
+function reducer(state, action) {
```
patrickkwang, line 11: It might be nice to name this something more informative. I'm not sure how broad the scope of this thing is, though. "selectionReducer"?

frontend/src/stores/useBiolinkModel.js
```diff
@@ -47,57 +53,57 @@ export default function useBiolinkModel() {
   }
 
   /**
-   * Get a list of ancestors for all biolink classes
+   * Get all descendants by getting direct descendants recursively
+   * @param {string} parentClass - biolink class to get direct descendants of
+   * @param {object} classes - object of all biolink classes
+   * @returns {string[]} list of all descendants
+   */
+  function getDirectDescendants(parentClass, classes) {
```
patrickkwang, line 61:
```suggestion
  function getDescendants(parentClass, classes) {
```

frontend/src/pages/answer/kgBubble/KgBubble.jsx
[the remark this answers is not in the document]
sarah-seitanakis: I agree that if the user is dragging a circle to the edge of the box, it would be more natural to see the bubble push into the edge a little. But, we don't want this behavior when the graph is resting-- only when the user is moving things around.

frontend/src/pages/answer/Answer.jsx
```diff
@@ -0,0 +1,124 @@
...
+
+  return (
+    <>
+      <LeftDrawer
+        displayState={displayState}
+        updateDisplayState={updateDisplayState}
+        onUpload={onUpload}
+      />
+      <div id="answerContentContainer">
+        <pageStatus.Display />
```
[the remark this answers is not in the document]
maximusunc: No. `pageStatus` is used on a bunch of pages for consistent loading and error messaging. `displayState` is specific to the answer page.

frontend/src/pages/answer/fullKg/simulation.worker.js
maximusunc: This might be specific to webpack as you can see from these [examples](https://webpack.js.org/loaders/worker-loader/#examples). It knows this is a web worker file from the `.worker.js` extension. In vanilla JS, you would call it as `const worker = new Worker(<path/to/worker/file>);`

frontend/src/pages/answer/kgBubble/KgBubble.jsx
maximusunc: Opened #193.

frontend/src/pages/answer/fullKg/simulation.worker.js
maximusunc: This might be specific to webpack (see [examples](https://webpack.js.org/loaders/worker-loader/#examples)) but webpack basically knows this is a web worker file and to auto-initialize it as a `Worker`. The same thing in vanilla javascript would be `const worker = new Worker(<path/to/webworkerfile.js>);`.

frontend/src/pages/answer/resultsTable/SupportingPublications.jsx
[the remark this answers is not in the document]
maximusunc: Added a no support publications message in 0acc1463b60388973064222b4180ed212a4c7faf

frontend/src/pages/answer/fullKg/KgFull.jsx
maximusunc: Fixed in b6653ab66a4515b3d14226e2f9311c169f3ea0d1

frontend/src/pages/answer/Answer.jsx
```diff
@@ -0,0 +1,124 @@
...
+        try {
+          msg = JSON.parse(fileContents);
+        } catch (err) {
+          displayAlert('error', 'Failed to read this message. Are you sure this is valid JSON?');
+        }
+        const errors = trapiUtils.validateMessage(msg);
+        if (!errors.length) {
+          try {
+            const standardQueryGraph = queryGraphUtils.standardize(msg.message.query_graph);
+            msg.message.query_graph = standardQueryGraph;
```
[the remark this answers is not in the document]
maximusunc: Fixed in 6371dc44d25ff3df4773de0e1f46adab099d0421

frontend/src/pages/answer/fullKg/KgFull.jsx
maximusunc: Fixed in 0c9e41ab05fd952e9b5cf15c79cc8db9aacea9a7

frontend/src/pages/answer/fullKg/__mocks__/simulation.worker.js
maximusunc: Updated in 0bf32c925c56cdabe83847a950f3ca76781ff742

frontend/src/pages/answer/fullKg/simulation.worker.js
maximusunc: Added in b45c753b1a59f99d276aca61585e4c6c2fca9bb9

frontend/src/pages/answer/resultsTable/SupportingPublications.jsx
maximusunc: Updated in b05f55ac3910ed20a5caa0ba0c5745bdcbe4f879

frontend/src/stores/useBiolinkModel.js
maximusunc: Fixed in a911937f47e64394d40209e43c9462292db59476

frontend/src/pages/answer/resultsTable/SupportingPublications.jsx
patrickkwang: Is there anything about this that is specific to publications? The name should describe what it does, as generally as possible. If it's used in the context of publications, so be it.

frontend/src/pages/answer/resultsTable/ResultExplorer.jsx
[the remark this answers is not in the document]
maximusunc: Fixed in 3531c7f8aa34bc2d48844ea6ba4988bdd4b574c2

frontend/src/pages/answer/kgBubble/KgBubble.jsx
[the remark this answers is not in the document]
maximusunc: Fixed in 1e02548abe4bb6558231f50c649df419087630e6

frontend/src/pages/answer/kgBubble/KgBubble.jsx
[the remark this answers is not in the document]
maximusunc: Updated in c53e41393655603c8f82d949f84ca8a652674352. The dragged node can node push into the edges a little while all other nodes are bound inside. Once you let go of the node, it is again constricted to inside.

frontend/src/pages/answer/resultsTable/SupportingPublications.jsx
maximusunc: Changed in aeb9adac67c32ad6ef20d4e54e233dc2976d88f7

frontend/src/pages/answer/Answer.jsx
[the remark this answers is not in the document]
patrickkwang: The fix looks good, but I want to consider options for displaying the KG without results. See #194.

frontend/src/pages/answer/fullKg/__mocks__/simulation.worker.js
patrickkwang: I feel like anywhere the mocked thing behaves differently from the actual thing, we should include documentation about how and why. Here my question is why `postMessage` does nothing by default. The answer may be tied in with how the tests work, but they live elsewhere.

frontend/src/pages/answer/fullKg/simulation.worker.js
patrickkwang: "create global electrostatic charge" doesn't really tell me what's going to happen. Is this attracting the nodes to each other? I'm also not sure what "add edges with length" means? Is it displaying the KG edges here?

maximusunc: Updated in f6124cbdc4c10c69c3af9d968c9e141558158dcc.

frontend/src/pages/answer/fullKg/__mocks__/simulation.worker.js
maximusunc: Updated in 0e1fb37bdd4814fcc3bc834e8c187eb42d9bcb6a

patrickkwang (approved):

PR #196, opened 2021-06-17
Implement the full new workflow from question creation to answer viewing

maximusunc:


frontend/src/utils/queryGraph.js
```diff
@@ -232,12 +244,27 @@ function prune(q_graph) {
   Object.keys(clonedQueryGraph.nodes).forEach((n) => {
     pruneEmptyArray(clonedQueryGraph.nodes[n], 'category');
     pruneEmptyArray(clonedQueryGraph.nodes[n], 'id');
-    if (clonedQueryGraph.nodes[n].id && clonedQueryGraph.nodes[n].category) {
+    // if (clonedQueryGraph.nodes[n].id && clonedQueryGraph.nodes[n].category) {
+    //   delete clonedQueryGraph.nodes[n].category;
```
sarah-seitanakis, line 248: Can these commented lines be removed?

frontend/tests/common/test_message.json
```diff
@@ -41,7 +41,7 @@
       {
         "node_bindings": {
           "n1": [{ "id": "1" }],
-          "n2": [{ "id": "2" }]
+          "n0": [{ "id": "2" }]
```
sarah-seitanakis, line 44: I know that this is just semantic and doesn't really affect anything, but wouldn't it make more sense if `n0` came before `n1` with the `id` of `n0` as 1 and the `id` of `n1` as 2? Just a thought.

query-dispatcher/routes.js
```diff
@@ -6,8 +6,9 @@ const { handleAxiosError } = require('./utils');
 
 // const messenger = process.env.MESSENGER_URL ||
 //                    'http://robokop.renci.org:4868/answer';
-const strider = process.env.STRIDER_URL ||
-                    'http://robokop.renci.org:5781';
+// const strider = process.env.STRIDER_URL ||
+//                     'http://robokop.renci.org:5781';
+const robokop = process.env.ROBOKOP_URL || 'http://robokop.renci.org:7092';
```
sarah-seitanakis, line 11: Is this referring to the external service RoboKop or is this for qgraph?

maximusunc: This is the external Robokop ARA.

sarah-seitanakis (changes requested):
You might want to go through and make sure that variables referring to the UI that contain `robokop` are changed to `qgraph`. I think it might be easier to catch these sooner, rather than later.

maximusunc:
> You might want to go through and make sure that variables referring to the UI that contain `robokop` are changed to `qgraph`. I think it might be easier to catch these sooner, rather than later.

Agreed. There will be a rebranding PR that will update all of our variables and text to reference 'qgraph'.

frontend/tests/common/test_message.json
maximusunc: Updated in 6354b586c0bbfa1213a8289bbdcce81e1e1816d9

frontend/src/utils/queryGraph.js
maximusunc: Removed in 74df36f5ada1baa2a9771ec61dfc50831e1aa063

sarah-seitanakis (approved):

frontend/tests/common/test_message.json
```diff
@@ -2,13 +2,13 @@
...
-          "id": "MONDO:0005737",
-          "is_set": false
-        },
         "n0": {
-          "category": "biolink:Gene",
+          "categories": ["biolink:Gene"],
           "is_set": true
+        },
+        "n1": {
+          "id": ["MONDO:0005737"],
```
patrickkwang, line 10: In TRAPI 1.1, this should be "ids".

patrickkwang (approved):
Looks good! I'll add some issues for future development.

frontend/tests/common/test_message.json
maximusunc: We talked offline. The UI is currently between TRAPI 1.0 and 1.1, and this is a side effect. There will be a future PR that will transition us completely to 1.1.

PR #205, opened 2021-07-09
Add ways to view results json

### LINKED ISSUE #198: View/download answer JSON

maximusunc:
- Add a download answer button to answer page
- Add a JSON view for each result in the Answer Explorer

Closes #198

### DIFF
--- a/frontend/src/pages/answer/resultsTable/SupportingPublications.jsx
+++ b/frontend/src/pages/answer/resultsTable/SupportingPublications.jsx
@@ -1,5 +1,8 @@
-import React, { useReducer, useEffect, useMemo } from 'react';
+import React, {
+  useState, useReducer, useEffect, useMemo,
+} from 'react';
 import Paper from '@material-ui/core/Paper';
+import IconButton from '@material-ui/core/IconButton';
 import List from '@material-ui/core/List';
 import ListItem from '@material-ui/core/ListItem';
 import ListItemText from '@material-ui/core/ListItemText';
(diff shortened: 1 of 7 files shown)


frontend/src/pages/answer/useAnswerStore.js
```diff
@@ -103,6 +110,8 @@ export default function useAnswerStore() {
       setSelectedResult({ nodes, edges });
       setSelectedRowId(rowId);
       setMetaData(edgePublications);
+      // store full result JSON
+      setResultJSON({ nodes: nodesJSON, edges: edgesJSON });
```
patrickkwang, line 114: I think we'd like to somehow include the result bindings, too. This is critical, for example, when the same node occurs twice in a single result.

patrickkwang (changes requested):

frontend/src/pages/answer/useAnswerStore.js
maximusunc: Since the nodes are a list in the JSON, is this an issue? If a node is repeated in the result bindings, it would show up twice in the JSON.

patrickkwang: Good point. It shouldn't. The value for `nodes` should be an object like in TRAPI, where the keys are CURIEs. Right now it looks like those ids are dropped. In addition, we need the bindings to know which knode is associated with which qnode. We're allowed, for example, to have multiple unpinned diseases in a question.

frontend/src/pages/answer/resultsTable/SupportingPublications.jsx
```diff
@@ -40,51 +45,71 @@ export default function SupportingPublications({ metaData }) {
...
+          <List>
+            <ListItem>
+              <ListItemText primary="No Supporting Publications Found" />
+            </ListItem>
+          </List>
+        )}
+      </div>
+      <div>
+        <h4>
+          Result JSON
```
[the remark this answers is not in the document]
maximusunc: Renamed in 22a2a9bc8724e430798a89ebe566234386f23186

frontend/src/pages/answer/useAnswerStore.js
maximusunc: Updated in 4cd6916

patrickkwang (approved):

PR #207, opened 2021-07-23
Add ability for multiple deployment options for different UI brands

maximusunc:


frontend/src/pages/Landing.jsx
```diff
@@ -88,7 +96,7 @@ export default function Landing() {
             <OptionButton
               glyph="import"
               header="Answer Explorer"
-              text="Easily upload JSON files of answersets to view them in Robokop&apos;s graphical interface."
+              text={`Easily upload JSON files of answersets to view them in ${brandConfig.title}'s graphical interface.`}
```
sarah-seitanakis, line 99: Does the apostrophe work okay here now? I saw that before you had to use \&apos;

frontend/src/robokop_config.json
```diff
@@ -0,0 +1,4 @@
+{
+    "brand": "robokop",
+    "title": "Robokop"
```
sarah-seitanakis, line 3: Just wanted to check that you meant to capitalize one (the title) and not the other!

sarah-seitanakis (changes requested):
A few of the files are flagged that there is no newline at the end of the file. Is that something that should be ignored? (See docker-compose.test.yml, frontend/src/qgraph_config.json, and frontend/src/robokop_config.json)

frontend/src/pages/Landing.jsx
maximusunc: Yeah. I think the apostrophe would have actually worked in the old format as well, the &apos might have been when that text was directly in the html. The new backtick allows normal apostrophes.

frontend/src/robokop_config.json
maximusunc: It's a little weird. I wanted the title to be capitalized because I'm putting that text directly into the page. The brand is more of a variable, so I didn't want it capitalized there. These config files are going to have a lot more in them once we start adding more styling differences between the two brands, but this is just the base logic.

maximusunc:
> A few of the files are flagged that there is no newline at the end of the file. Is that something that should be ignored? (See docker-compose.test.yml, frontend/src/qgraph_config.json, and frontend/src/robokop_config.json)

Added in bced189

sarah-seitanakis (dismissed):

frontend/src/robokop_config.json
patrickkwang: ROBOKOP is an acronym.
```suggestion
    "title": "ROBOKOP"
```

patrickkwang (approved):
Looks good! Minor fix.

patrickkwang (approved):

Issue #211, opened 2021-08-03
Valid TRAPI query unable to load


karafecho:
This issue is to report an error that is returned when an other valid (I think) JSON query is uploaded to the query graph builder.

![image](https://user-images.githubusercontent.com/26254388/128036488-1698e20b-6582-4bde-ba95-de24356d3093.png)

karafecho:
```
{
  "message": {
    "query_graph": {
      "nodes": {
        "n0": {
          "categories": [
            "biolink:GeneOrGeneProduct"
          ],
          "ids": ["NCBIGene:3105","NCBIGene:3106","NCBIGene:64167","NCBIGene:60412","NCBIGene:26191"]
        },
        "n1": {
          "categories": [
            "biolink:BiologicalProcessOrActivity"
          ]
        }
      },
      "edges": {
        "e0": {
          "predicates": [
            "biolink:has_participant"
          ],
          "subject": "n1",
          "object": "n0"
        }
      }
    }
  }
}
```

Issue #63, opened 2020-10-27
QuestionAnswerViewer should be able to display just a query graph


maximusunc:
If a question's results are invalid, we currently just display an error message saying we failed to parse the message. But we should be able to separate out the query graph from the results and display the graph if it's valid.

maximusunc:
I tried to upload a large message from the simple viewer. It created the question successfully because the query graph is valid, but then failed to upload the message because it was too large (will be fixed by #24). When I went to that question page, I got redirected to an answer page, and since the message upload failed, this answer has no data. We should be able to display the query graph here even though the answer display failed.

maximusunc:
This will happen in other instances too, like if the user deletes all answers for a question, or any time we fail to get an answer back. The user should be able to see what query graph was used.

PR #213, opened 2021-08-03
Change internal message state to be the full message format

### LINKED ISSUE #206: Display full TRAPI JSON message on Query Builder page
This issue is to modify the query builder such that it returns a full TRAPI JSON message to users, not just the body of the message.

maximusunc:
Closes #206, #195, #63, #211


patrickkwang (approved):
Looks good!

PR #225, opened 2021-09-02
Fix how brand is passed from docker to react

maximusunc:
The `environment` property wasn't passing the BRAND into the frontend container correctly, so `process.env.BRAND` was always undefined. Passing BRAND as an ARG into the Dockerfiles seems to do the trick.

### DIFF
--- a/docker-compose.dev.yml
+++ b/docker-compose.dev.yml
@@ -10,8 +10,8 @@ services:
   frontend:
     build:
       dockerfile: Dockerfile.dev
-    environment:
-      BRAND: qgraph
+      args:
+        BRAND: qgraph
     volumes:
       - ./frontend:/app
       # Mount this in an anonymous volume so that the locally available node_modules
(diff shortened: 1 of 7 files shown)


alongreyber (approved):
Ok I see that this is necessary because `process.env` is evaluated at build time rather than runtime. This seems like a good solution to use build args. 

One possible alternative solution is [this one](https://blog.freemiumpn.com/docs/developer/how-to/web/runtime-configs-with-webpack-and-docker-v2/), where `process.env` is redefined so that it's updated dynamically at runtime. The advantage of solution 2 is that it could work better if we needed access to more env variables from the frontend in the future. I'll leave it up to you which we go with. It's also super simple (4 lines of code) which is nice.

maximusunc:
For now, I don't foresee having to add more environment variables to the frontend like this, so I'm going to go ahead and merge. I'll keep this other solution in mind.

PR #226, opened 2021-09-07
Enable Robokache features through the UI

maximusunc:
Enable
- saving answers
- deleting answers
- deleting questions and associated answers
- setting question visibility

### DIFF
--- a/Robokache_API.md
+++ b/Robokache_API.md
@@ -2,7 +2,7 @@
 
 1. You need to have an email/password account through the qgraph UI. Please visit https://qgraph.org and create an account.
 
-2. The following cURL will validate your user and provide you an access token that you can use with Robokache
+2. The following cURL will validate your user and provide you an access token that you can use with Robokache. The `client_id` is specific to Auth0 and you shouldn't have to do anything with it.
 ```bash
 curl --request POST \
   --url 'https://qgraph.us.auth0.com/oauth/token' \
--- a/Robokache_API.md
+++ b/Robokache_API.md
@@ -1,0 +1,32 @@
+# How to interact with Robokache through its API
+
+1. You need to have an email/password account through the qgraph UI. Please visit https://qgraph.org and create an account.
+
+2. The following cURL will validate your user and provide you an access token that you can use with Robokache
+```bash
+curl --request POST \
+  --url 'https://qgraph.us.auth0.com/oauth/token' \
+  --header 'content-type: application/x-www-form-urlencoded' \
+  --data-urlencode grant_type=password \
+  --data-urlencode username={email} \
+  --data-urlencode password={password} \
+  --data-urlencode 'client_id=sgJrK1gGAbzrXwUp0WG7jAV0ivCIF6jr'
+```
+
+3. You then need to create an answer document on Robokache with metadata which will return you an answer_id
+```bash
+curl --request POST \
+  --url 'https://qgraph.org/api/robokache/document' \
+  --header 'Authorization: Bearer {access_token}' \
+  --data '{"parent":"","visibility":2,"metadata":{"name":"Uploaded Answer","answerOnly":true,"hasAnswers":true}}'
+```
+
+4. Once an answer has been created, you can upload your message using the answer_id
+```bash
+curl --request PUT \
+  --url 'https://qgraph.org/api/robokache/document/{answer_id}/data' \
+  --header 'Authorization: Bearer {access_token}' \
+  --data '{message}'
+```
+
+5. To view this message in the UI, go to https://qgraph.org/answer/{answer_id}
--- a/frontend/src/pages/answer/Answer.jsx
+++ b/frontend/src/pages/answer/Answer.jsx
@@ -116,6 +116,8 @@ export default function Answer() {
 
   /**
    * Get metadata of answer to see if current user owns it
+   *
+   * Disable the "Delete Answer" button if answer is not owned.
    */
   async function checkIfAnswerOwned() {
     let accessToken;
--- a/frontend/src/pages/answer/Answer.jsx
+++ b/frontend/src/pages/answer/Answer.jsx
@@ -106,7 +106,7 @@ export default function Answer() {
       try {
         accessToken = await getAccessTokenSilently();
       } catch (err) {
-        displayAlert('error', `Failed to validate user. Error: ${err}`);
+        displayAlert('error', `Failed to authenticate user: ${err}`);
       }
     }
     const answerResponse = await API.cache.getAnswerData(answer_id, accessToken);
@@ -125,7 +125,7 @@ export default function Answer() {
       try {
         accessToken = await getAccessTokenSilently();
       } catch (err) {
-        displayAlert('error', `Failed to validate user. Error: ${err}`);
+        displayAlert('error', `Failed to authenticate user: ${err}`);
       }
     }
     const answerResponse = await API.cache.getAnswer(answer_id, accessToken);
@@ -227,19 +227,19 @@ export default function Answer() {
       try {
         accessToken = await getAccessTokenSilently();
       } catch (err) {
-        displayAlert('error', `Failed to validate user. Error: ${err}`);
+        displayAlert('error', `Failed to authenticate user: ${err}`);
       }
     } else {
       return displayAlert('warning', 'You need to be signed in to save an answer.');
     }
     let response = await API.cache.createAnswer(defaultAnswer, accessToken);
     if (response.status === 'error') {
-      return displayAlert('error', 'Failed to create answer.');
+      return displayAlert('error', `Failed to create answer: ${response.message}`);
     }
     const answerId = response.id;
     response = await API.cache.setAnswerData(answerId, { message: answerStore.message }, accessToken);
     if (response.status === 'error') {
-      return displayAlert('error', 'Failed to save answer.');
+      return displayAlert('error', `Failed to save answer: ${response.message}`);
     }
     return displayAlert('success', 'Your answer has been saved!');
   }
@@ -256,7 +256,7 @@ export default function Answer() {
       try {
         accessToken = await getAccessTokenSilently();
       } catch (err) {
-        displayAlert('error', `Failed to validate user. Error: ${err}`);
+        displayAlert('error', `Failed to authenticate user: ${err}`);
       }
     } else {
       return displayAlert('warning', 'You need to be signed in to delete an answer.');
--- a/frontend/src/pages/answer/Answer.jsx
+++ b/frontend/src/pages/answer/Answer.jsx
@@ -1,4 +1,6 @@
-import React, { useEffect, useContext, useMemo } from 'react';
+import React, {
+  useState, useEffect, useContext, useMemo,
+} from 'react';
 import { useRouteMatch, useHistory } from 'react-router-dom';
 import { get as idbGet, del as idbDelete, set as idbSet } from 'idb-keyval';
 import { useAuth0 } from '@auth0/auth0-react';
@@ -11,6 +13,7 @@ import queryGraphUtils from '~/utils/queryGraph';
 
 import useAnswerStore from './useAnswerStore';
 import useDisplayState from './useDisplayState';
+import { defaultAnswer } from '~/utils/cache';
 
 import LeftDrawer from './leftDrawer/LeftDrawer';
 import KgBubble from './kgBubble/KgBubble';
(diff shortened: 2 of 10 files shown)


sarah-seitanakis (approved):
Looks good from the UI side!

alongreyber (approved):

frontend/src/pages/questionList/subComponents/QuestionTableRow.jsx
```diff
@@ -1,24 +1,54 @@
...
   const visibility = useVisibility();
   const displayAlert = useContext(AlertContext);
   const { isAuthenticated, getAccessTokenSilently } = useAuth0();
+  const [confirmOpen, setConfirmOpen] = useState(false);
+
+  /**
+   * Update question visibility on Robokache
+   * @param {boolean} checked - is the 'public' switch on
+   */
+  async function setVisibility(checked) {
```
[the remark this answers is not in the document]
maximusunc: Yeah, it's toggling between public and shareable. @sarah-seitanakis and I discussed and for now, all messages are shareable by default or you can make it public and it'll show up in the public library.

frontend/src/pages/answer/Answer.jsx
```diff
@@ -192,13 +216,98 @@ export default function Answer() {
     event.target.value = '';
   }
 
+  /**
+   * Save an uploaded answer to Robokache
+   */
+  async function saveAnswer() {
```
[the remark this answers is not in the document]
maximusunc: I'm not sure I understand. The "Upload Answer" button is for stand-alone messages. They have no associated question (or "parent" document). Are you saying you have a message where the `query_graph` is not matched with its `results` or `knowledge_graph`?

frontend/src/pages/questionList/subComponents/QuestionTableRow.jsx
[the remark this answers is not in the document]
alongreyber: Shouldn't the word "Public" change to the word "Shareable" when I switch the visibility?

frontend/src/pages/answer/Answer.jsx
```diff
@@ -192,13 +216,98 @@ export default function Answer() {
...
+    let accessToken;
```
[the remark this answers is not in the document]
maximusunc: I'm not sure how you got here. Could you provide steps to reproduce? Taking a look, the one question you have was asked from the question page and it looks like the ARA failed to return a message, so it shows it has no answers. If you upload an answer on the answer page, its name would be "Uploaded Answer".

frontend/src/pages/answer/Answer.jsx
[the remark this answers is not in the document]
alongreyber: I assumed looking at the UI that the "Upload Answer" button on the left means "Upload a new answer to the question that I am currently viewing." Is that not correct?

frontend/src/pages/questionList/subComponents/QuestionTableRow.jsx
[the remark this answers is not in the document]
sarah-seitanakis: I still think it should always be shareable. so whether public is off or on, shareable is always on. the decision for the user is whether they should make their question public or not public.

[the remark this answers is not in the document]
sarah-seitanakis: we can gray the word "Public" when it is switched off, I think that would be good

frontend/src/pages/answer/Answer.jsx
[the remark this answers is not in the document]
maximusunc: No, "Upload Answer" is a simple viewer where you can view a stand-alone message. Maybe there should be a confirm dialog saying what's going to happen when you upload an answer. wrt your assumption, I don't think we should allow users to upload their own answers to existing questions, but that they would have to re-ask the question through the UI.

frontend/src/pages/answer/Answer.jsx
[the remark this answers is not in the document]
maximusunc: Either the ARA failed, or you went to the answer page and deleted the only answer, so now the question has no answers.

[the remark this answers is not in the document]
maximusunc: Right now there's no way to go re-ask a question, but that's next on my list.

maximusunc:
How about we make the column header say "Public" with switches in the rows?
Screen Shot 2021-09-13 at 11 18 28 AM

sarah-seitanakis (dismissed):

frontend/src/pages/answer/Answer.jsx
```diff
@@ -111,16 +114,37 @@ export default function Answer() {
...
+  /**
+   * Get metadata of answer to see if current user owns it
+   */
+  async function checkIfAnswerOwned() {
+    let accessToken;
+    if (isAuthenticated) {
+      try {
+        accessToken = await getAccessTokenSilently();
+      } catch (err) {
+        displayAlert('error', `Failed to validate user. Error: ${err}`);
```
patrickkwang, line 126:
```suggestion
        displayAlert('error', `Failed to authenticate user. Error: ${err}`);
```

frontend/src/pages/answer/Answer.jsx
```diff
@@ -111,16 +114,37 @@ export default function Answer() {
...
+      }
+    }
+    const answerResponse = await API.cache.getAnswer(answer_id, accessToken);
```
patrickkwang, line 129: Is determining whether the answer is owned really the only reason we need the metadata? It seems odd that we make the API call here, grab `.owned`, and throw away the rest.

frontend/src/pages/answer/Answer.jsx
```diff
@@ -192,13 +216,98 @@ export default function Answer() {
...
+  /**
+   * Save an uploaded answer to Robokache
+   */
+  async function saveAnswer() {
+    let accessToken;
+    if (isAuthenticated) {
+      try {
+        accessToken = await getAccessTokenSilently();
+      } catch (err) {
+        displayAlert('error', `Failed to validate user. Error: ${err}`);
```
patrickkwang, line 228:
```suggestion
        displayAlert('error', `Failed to authenticate user. Error: ${err}`);
```

frontend/src/pages/answer/Answer.jsx
```diff
@@ -192,13 +216,98 @@ export default function Answer() {
...
+      }
+    } else {
+      return displayAlert('warning', 'You need to be signed in to save an answer.');
+    }
+    let response = await API.cache.createAnswer(defaultAnswer, accessToken);
+    if (response.status === 'error') {
+      return displayAlert('error', 'Failed to create answer.');
```
patrickkwang, line 235: Can we pass along any details to the user re: why we failed?

frontend/src/pages/answer/Answer.jsx
```diff
@@ -192,13 +216,98 @@ export default function Answer() {
...
+   * Also check if the associated question has other answers
+   * and set `hasAnswers` to false if it doesn't
+   */
+  async function deleteAnswer() {
+    let accessToken;
+    if (isAuthenticated) {
+      try {
+        accessToken = await getAccessTokenSilently();
+      } catch (err) {
+        displayAlert('error', `Failed to validate user. Error: ${err}`);
```
patrickkwang, line 257:
```suggestion
        displayAlert('error', `Failed to authenticate user. Error: ${err}`);
```

frontend/src/pages/answer/Answer.jsx
```diff
@@ -192,13 +216,98 @@ export default function Answer() {
...
+        alertText += ` However, failed to get sibling answers: ${response.message}`;
+      } else if (!response.length) {
+        // if question has no answers, set hasAnswers to false
+        response = await API.cache.getQuestion(questionId, accessToken);
+        if (response.status === 'error') {
+          alertType = 'error';
+          alertText += ` However, failed to get associated question data: ${response.message}`;
+        } else {
+          response.metadata.hasAnswers = false;
+          response = await API.cache.updateQuestion(response, accessToken);
```
patrickkwang, line 290: It seems odd that this is the UI's job. Also, would it make more sense for hasAnswers to be `false` by default, and to set it to `true` only after we get answers?

frontend/src/pages/questionList/subComponents/QuestionTableRow.jsx
```diff
@@ -32,19 +62,59 @@ export default function QuestionTableRow({ question }) {
...
+    }
+  }
+
+  async function deleteQuestion() {
+    let accessToken;
+    if (isAuthenticated) {
+      try {
+        accessToken = await getAccessTokenSilently();
+      } catch (err) {
+        displayAlert('error', `Failed to get user questions. Error: ${err}`);
```
patrickkwang, line 93:
```suggestion
        displayAlert('error', `Failed to get authenticate user. Error: ${err}`);
```

patrickkwang (changes requested):
Some questions.

Robokache_API.md
```diff
@@ -0,0 +1,32 @@
...
+
+2. The following cURL will validate your user and provide you an access token that you can use with Robokache
+```bash
+curl --request POST \
+  --url 'https://qgraph.us.auth0.com/oauth/token' \
+  --header 'content-type: application/x-www-form-urlencoded' \
+  --data-urlencode grant_type=password \
+  --data-urlencode username={email} \
+  --data-urlencode password={password} \
+  --data-urlencode 'client_id=sgJrK1gGAbzrXwUp0WG7jAV0ivCIF6jr'
```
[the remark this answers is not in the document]
maximusunc: This is the UID for our application on Auth0. It needs to be passed in this call so Auth0 knows which application to authenticate against. The user doesn't need to do anything with it. I can add something to that instruction explaining what it is and not to touch it.

frontend/src/pages/answer/Answer.jsx
maximusunc: Yes. This call is very quick, as we only get the metadata of the answer, not the actual message. This is so we can disable the "Delete Answer" button so non-answer owners don't think they can even delete an answer they don't own.

frontend/src/pages/answer/Answer.jsx
maximusunc: This function is working on deleting an answer. Right here, we've just determined that we deleted all answers. So the question had `hasAnswers=true`, now we need to set it to `false` since we just deleted the last answer. I think the reason the UI has to do this was decided in Robokache, because we wanted it to be ignorant of any metadata and data it is storing. As generic document storage, it shouldn't require that metadata have a specific property.

Robokache_API.md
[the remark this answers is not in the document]
maximusunc: Added in df5308d6a1e26386aba4ef0b897106f9e0a2c2d2

frontend/src/pages/answer/Answer.jsx
maximusunc: Add documentation in df5308d6a1e26386aba4ef0b897106f9e0a2c2d2

frontend/src/pages/answer/Answer.jsx
maximusunc: Updated in 06dfb063189bcbd4f9ce23d3ace08fba1c9301f6

frontend/src/pages/answer/Answer.jsx
maximusunc: Updated in 06dfb063189bcbd4f9ce23d3ace08fba1c9301f6

frontend/src/pages/answer/Answer.jsx
maximusunc: Updated in 06dfb063189bcbd4f9ce23d3ace08fba1c9301f6

frontend/src/pages/answer/Answer.jsx
maximusunc: Updated in 06dfb063189bcbd4f9ce23d3ace08fba1c9301f6

frontend/src/pages/questionList/subComponents/QuestionTableRow.jsx
maximusunc: Updated in 06dfb063189bcbd4f9ce23d3ace08fba1c9301f6

frontend/src/pages/answer/Answer.jsx
patrickkwang: Let's keep an eye on this. I can imagine us wanting a variety of different bits of metadata, and we don't want to end up making the same call repeatedly.

frontend/src/pages/answer/Answer.jsx
patrickkwang: I think we should rethink this `hasAnswers` business, but let's roll with it for now.

patrickkwang (approved):

Issue #222, opened 2021-08-30
Fix question page redirect when we get an answer back


maximusunc:
The question page was at "/question", but now it can either be "/" or "/question" based on branding. We need to handle either case when determining if the question page should redirect to the answer page or not.

Issue #220, opened 2021-08-26
Display more answers per page


karafecho:
This issue is to suggest that QGraph increases the number of answers it displays per page, e.g., 50 or 100.

Issue #218, opened 2021-08-26
Display ranked scores


karafecho:
This issue is to suggest that QGraph displays ranked scores in the tabular view.

Issue #194, opened 2021-06-08
Display KG cards when results are empty


patrickkwang:
From https://github.com/NCATS-Gamma/robokop-rewrite/pull/187#discussion_r645645504

PR #228, opened 2021-09-16
Misc updates

maximusunc:
This PR fixes a lot of issues:
- #31 
- #109 
- #144 
- #148 
- #155 
- #192 
- #194 
- #218 
- #220 
- #222


frontend/src/utils/results.js
```diff
@@ -109,8 +109,19 @@ function makeTableHeaders(message, colorMap) {
...
         return knowledge_graph.nodes[value[0].id].name || value[0].id;
       },
+      disableSortBy: true,
     };
   });
+  if (results[0].score) {
+    const scoreColumn = {
+      Header: 'Score',
+      id: 'score',
+      accessor: (row) => Math.round(row.score * 1000) / 1000,
```
patrickkwang, line 140: What assumption(s) is this making about the range of the scores?

frontend/src/pages/queryBuilder/textEditor/textEditorRow/NodeSelector.jsx
```diff
@@ -115,6 +123,15 @@ export default function NodeSelector({
     }
   }, [open, searchTerm]);
 
+  /**
+   * Cancel any api calls on unmount
```
patrickkwang, line 127: I'm not clear on when this is supposed to happen. Does the component get unmounted whenever the user changes what is typed in the search box?

frontend/src/utils/results.js
maximusunc: This will round scores to the 1/1000th. I was getting scores from Robokop that had quite the number of decimal places. It didn't look very good. But we can do it smarter if needed. I'm assuming scores can vary between ARAs, some being between 0-1, others being well over 1?

frontend/src/pages/queryBuilder/textEditor/textEditorRow/NodeSelector.jsx
maximusunc: No, this happens when the user navigates to a different page.

frontend/src/utils/results.js
patrickkwang: If the goal is to limit the display length, we might want to put a limit on total digits instead somehow, but this will work for now.

frontend/src/pages/queryBuilder/textEditor/textEditorRow/NodeSelector.jsx
patrickkwang: Ah, I was confused, the issue is solved at line 103.

patrickkwang (approved):
The code looks good to me. I have not done any actual testing.

sarah-seitanakis (approved):

PR #232, opened 2021-10-01
Answer viewer fixes

maximusunc:
- Add pruning to the kg bubble graph
- Add pruning to the answer explorer graph
- Handle null kg and results in messages
- Update how we pick node category colors


frontend/src/utils/colors.js
```diff
@@ -15,7 +15,6 @@ const conceptColorMap = {
   'biolink:Disease': '#fbb4ae', // Red
   'biolink:DiseaseOrPhenotypicFeature': '#fbb4ae', // Red, same as disease
   'biolink:Drug': '#8787ff', // Purply blue, same as chemical_substance
-  'biolink:Protein': '#ccebc5', // Green like gene
   'biolink:EnvironmentalFeature': '#8a9a5b', // Moss Green
   'biolink:Food': '#ffa343', // Neon Carrot
```
patrickkwang, line 19: Good band name.

sarah-seitanakis:
image
The color of the middle node in the query graph and the table not matching the bubble graph and answer explorer graph is kind of confusing. I do see that the answer I selected has a middle node with category "small molecule". Since this is a sub category of the broader "molecular entity" can we make it a hue of that color? Or is there another way to make this less confusing?

sarah-seitanakis:
image
Is there a way that we not have a pruning slider on graphs that will not require pruning? I don't like that these graphs have the pruning slider since a three node answer will not need it, nor will a bubble graph like this.

sarah-seitanakis:
image
It looks like this answer has a publication but it is not showing in the answer explorer

sarah-seitanakis (changes requested):
There are a couple things that I commented about

maximusunc:
@sarah-seitanakis
- I updated the color mappings. Now we aren't specifying a color for small molecule, so nodes will go with the color of their closest parent category's color.
- I removed the prune slider for smaller graphs.
- The edge publication attribute format is not standardized, so I'm seeing new formats occasionally. I've updated the publication logic to include all the formats I'm aware of currently.

sarah-seitanakis (changes requested):
image
Thanks for the fixes, it's looking great! One more thing-- do you think you could add some padding to the text that I highlighted in the answer explorer? Also, for some reason, the publication is still not showing up

maximusunc:
@sarah-seitanakis That's a publication on a specific node, which I don't think we've ever handled? We've only ever showed publications for edges. @patrickkwang should we be showing links for these node publications?

patrickkwang:
> should we be showing links for these node publications?

No, I'm not sure what those would be for. Unless someone asks for them, let's ignore.

sarah-seitanakis (approved):
Looks great!!

sarah-seitanakis (approved):

patrickkwang (approved):
Looks great!

PR #238, opened 2022-05-04
Remove excess docker containers

maximusunc:
Removes proxy and query-dispatcher and just leaves qgraph (frontend) and robokache. qgraph is now an express server that serves the html as well as handles all api calls.


src/pages/queryBuilder/QueryBuilder.jsx
```diff
@@ -41,9 +41,8 @@ export default function QueryBuilder() {
    */
   async function onQuickSubmit() {
     pageStatus.setLoading('Fetching answer, this may take a while');
-    const ara_url = ARAs[ara];
     const prunedQueryGraph = queryGraphUtils.prune(queryBuilder.query_graph);
-    const response = await API.ara.getAnswer(ara_url, { message: { query_graph: prunedQueryGraph } });
+    const response = await API.ara.getQuickAnswer(ara, { message: { query_graph: prunedQueryGraph } });
```
kennethmorton, line 45: Double checking that this should now be getQuickAnswer and not getAnswer?

kennethmorton, line 45: Double checking that this should now be getQuickAnswer and not getAnswer?

kennethmorton (approved):
There is a lot going on here. This seems to cover more directions than just removing the proxy container. There is also a decent number of backend UI server changes. Given the volume of changes and the visual quality of the code, I am content to settle for testing rather than rigorous review. We may find a few bugs once we get more miles on this code, but finding them in the wild will be easier than in this review process.

src/pages/queryBuilder/QueryBuilder.jsx
maximusunc: Yeah, this one gets sent directly to the ara without getting saved in robokache. Line 73 sends the question id instead of the message because it's interacting with robokache.

PR #246, opened 2023-03-20
prerelease

cbizon:
* made robokop the only option
* updated urls in dockerfile
* removed filtering from the chooser

### DIFF
--- a/src/utils/fetchCuries.js
+++ b/src/utils/fetchCuries.js
@@ -2,7 +2,6 @@ import API from '~/API';
 
 export default async function fetchCuries(entity, displayAlert, cancel) {
   // Get list of curies that match this search term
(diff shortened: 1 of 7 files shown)


src/utils/fetchCuries.js
```diff
@@ -2,6 +2,7 @@ import API from '~/API';
 
 export default async function fetchCuries(entity, displayAlert, cancel) {
   // Get list of curies that match this search term
+  console.log('lookup ${entity}');
```
maximusunc, line 5: If you want to include the actual entity variable in this log message, I believe you need to use backticks here instead of single quotes.

cbizon: Yeah, didn't mean to commit those console messages anyway.

PR #249, opened 2023-03-23
update funding message

Woozl:
Updates the abbreviated funding message to:

> ROBOKOP is a joint creation of RENCI and CoVar. Early development was supported by the NCATS and continued development is supported by the NIEHS and ODSS within the NIH.

There's also this option:

> ROBOKOP is a joint creation of RENCI and CoVar, which received initial support from NCATS and is now supported by NIEHS and ODSS within the NIH.

Let me know if this is good or if you'd like it adjusted


karafecho (approved):
Looks good to me! Thanks, David.

PR #250, opened 2023-03-27
Update/edge properties

Woozl:
https://user-images.githubusercontent.com/16181779/227975721-2c037ae9-22ff-4180-8b28-a4be482ad644.mov

For now, I've simply produced a table with all the attribute key-value pairs. If we'd like, I can change this to just show the `primary_knowledge_source`, and maybe link out to the sources. Let me know what you think.


mbwatson:
this is a really good start!

cbizon (approved):

PR #254, opened 2023-05-04
Fix predicate selector

Woozl:
The predicate selector now considers the subject's/object's (and their ancestor's) mixins. It no longer considers the subject's/object's descendant classes


mbwatson (approved):
:+1: looks good!

PR #266, opened 2023-07-31
Parse TRAPI 1.4

Woozl:
Parses TRAPI updated `result` object. All analysis edges are aggregated and scores averaged per result.

This parsing is very barebones, just to get it functioning with the current UI. We should probably discuss a UI that provides more fine-grained information using support graphs and individual analyses.


Woozl:
This is tagged/released as [v1.1.5-beta-4](https://github.com/RobokopU24/qgraph/releases/tag/v1.1.5-beta-4), but I'm going to go ahead and merge to the working branch here as it's now used in production

PR #269, opened 2023-11-15
Bug uichanges

hina-shah:
This PR fixes two bugs:
- 0 was showing up when answer explorer was expanded on the Answer page
- On the answer page, footer stayed behind the left drawer. 

An update to us html's semantic <footer> tag instead of <div> has also been made.

### DIFF
--- a/src/pages/answer/resultsTable/ResultExplorer.jsx
+++ b/src/pages/answer/resultsTable/ResultExplorer.jsx
@@ -350,7 +350,7 @@ export default function ResultExplorer({ answerStore }) {
       elevation={3}
     >
       <h5 className="cardLabel">Answer Explorer</h5>
-      { Boolean(answerStore.showNodePruneSlider)  && (
+      { Boolean(answerStore.showNodePruneSlider) && (
         <Box width={200} id="nodeNumSlider">
           <Slider
             value={numTrimmedNodes}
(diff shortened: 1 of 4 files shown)


src/pages/answer/resultsTable/ResultExplorer.jsx
```diff
@@ -350,7 +350,7 @@ export default function ResultExplorer({ answerStore }) {
       elevation={3}
     >
       <h5 className="cardLabel">Answer Explorer</h5>
-      {answerStore.showNodePruneSlider && (
+      { Boolean(answerStore.showNodePruneSlider)  && (
```
Woozl, line 353: Actually, I just saw, this in the CI/CD lint checker. Looks like there are two spaces here:

image

You can run `npm run lint` to check the whole project before pushing, otherwise it won't let us merge 🙄. Also, install the "ESLint" plugin in VSCode so it will highlight these issues for you. To fix this one, just submit another commit and it will rerun the CI/CD.

Woozl (changes requested):

src/pages/answer/resultsTable/ResultExplorer.jsx
hina-shah: Ok great! Yeah I was confused about this. Thanks!

Woozl (approved):

PR #275, opened 2024-03-19
Update/remove nodenorm

Woozl:


Woozl:
This was originally created to support the yobokop UI, which has a lot of nodes not in babel, so they returned `null` in the call to nodenorm. After discussing, it doesn't seem that we need the call to nodenorm in the main UI either, as calls to nameres will always have the preferred identifier.

PR #301, opened 2025-03-18
FEAT: Adding filtering to the each column of the results table

hina-shah:
This PR adds a filtering option to the columns in the results table to select from the options in the column.


Woozl (approved):
👍

PR #307, opened 2025-05-21
Bug fixes symmetric predicates/edges showing arrows (#257)

Amaru333:
Fixed bug where symmetrical predicates/edges were displaying arrows instead of solid lines (Issue #257)

image

image

### DIFF
--- a/src/utils/d3/graph.js
+++ b/src/utils/d3/graph.js
@@ -215,8 +215,8 @@ function isInside(x, y, cx, cy, r) {
  * @param {obj} edge edge object
  * @returns {str} url(#arrow) or empty string
  */
-function shouldShowArrow(edge) {
-  return (edge.predicates && edge.predicates.findIndex((p) => p !== 'biolink:related_to') > -1) || (edge.predicate && edge.predicate !== 'biolink:related_to');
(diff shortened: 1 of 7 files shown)


src/utils/d3/graph.js
```diff
@@ -215,8 +215,8 @@ function isInside(x, y, cx, cy, r) {
  * @param {obj} edge edge object
  * @returns {str} url(#arrow) or empty string
  */
-function shouldShowArrow(edge) {
-  return (edge.predicates && edge.predicates.findIndex((p) => p !== 'biolink:related_to') > -1) || (edge.predicate && edge.predicate !== 'biolink:related_to');
+function shouldShowArrow(edge, symmetric_predicates = ['biolink:related_to']) {
```
Woozl: Could we name the parameter `symmetricPredicates` to match the other variables?

Woozl (changes requested):
Looks great! Great job getting this working quickly. I just suggested one small style fix to follow the camelCase naming conventions. 

It's a bit confusing because there are some places in the app that read from external data sources that are written in Python, where the convention is to use snake_case. But for these variables we're creating in Javascript, lets stick to the convention.

Woozl (approved):
👍

PR #308, opened 2025-06-10
Bugfix/hot reload

Amaru333:
- Fixed hot reload issue.
- `npm run dev` now uses hot reload.
- React app runs in port 8080 (http://localhost:8080) while the nodejs server runs in port 7080.
- Added a proxy to redirect all the API calls to port 7080 in the dev webpack

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,6 +44,8 @@ You may also directly invoke docker-compose with the following command:
 docker-compose -f docker-compose.base.yml -f docker-compose.dev.yml up --build
 ```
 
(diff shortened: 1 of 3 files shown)


Woozl (approved):
This is a great fix and a huge improvement, thanks for handling this

Woozl:
The only thing I noticed was the `localhost:7080` port still responds with the non-reloading version, since the `server.js` also points to the app:

https://github.com/RobokopU24/qgraph/blob/14db2d934ec0eedb369d6ae289f7d1c9582cf2e1/server.js#L32-L36

Can we put a brief note in the README saying to use localhost:8080? @Amaru333

